PR #8, opened 2025-02-13
Introduce in-memory store for tenant configurations, refactor and rename

iggy-da:
Introduces a new repository to store meta-data associated with tenant registrations in-memory


quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
```diff
@@ -16,25 +20,55 @@
...
+        List<String> authorities = auth.getAuthorities()
+                .stream()
+                .map(GrantedAuthority::getAuthority)
+                .toList();
+
+        // Retrieve the registrationId from the authentication
+        String registrationId = auth.getAuthorizedClientRegistrationId();
+
+        // Default walletUrl fallback if not found
+        String walletUrl = "http://localhost/";
```
peterkvokacka-da, line 49: If we don't have a reasonable default why not keep it null ?

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
```diff
@@ -16,25 +20,55 @@
...
                 party.split("::")[0],
+                // party
                 party,
+                // roles
                 authorities,
-                authorities.contains("ROLE_ADMIN")
+                // isAdmin
+                authorities.contains("ROLE_ADMIN"),
+                // walletUrl
+                walletUrl
```
peterkvokacka-da, line 68: I would probably opted for tenantId here instead of walletUrl as I don't think AuthenticatedUser should hold every possible tenant property we may need in the future.

peterkvokacka-da (changes requested):

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AdminApiImpl.java
```diff
@@ -3,34 +3,48 @@
...
+public class AdminApiImpl implements AdminApi {
 
+    private final OAuth2ClientRegistrationRepository tenantRegistrationRepository;
+    private final TenantPropertiesRepository tenantPropertiesRepository;
 
     @Autowired
-    public AdminApiImpl(OAuth2ClientRegistrationRepository clientRegistrationRepository) {
-        this.clientRegistrationRepository = clientRegistrationRepository;
+    public AdminApiImpl(
+            OAuth2ClientRegistrationRepository tenantRegistrationRepository,
```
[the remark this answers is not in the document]
iggy-da: > It is likely a good starting point for the most customers

Agree, and I think then it fulfills its purpose. Any serious app will have to rewrite the implementation with something that is not in-memory anyway.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
iggy-da: Good point

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
iggy-da: Base on this feedback I started to go down this road but I don't think it makes the code any better - rather the opposite.

The logged in user can only query information about themselves, not other tenants. A by-tenant-id join does not make sense here.

Adding another `/me`-style endpoint also does not add any value. Just return the information for the logged in user. This is conventional and splitting the information into multiple endpoints just adds unnecessary complexity.

Let me know what you think.

peterkvokacka-da: It should be rather straight forward: 
```
String tenantId = getAuthenticatedUser().getTenantId()
tenantPropertiesRepository.getTenant(tenantId).getWalletUrl()
``` 
if for some reason it is problematic I can live with walletUrl on AuthenticatedUser for now and have look at this later when I am done with Authentication rework. 

Please do not introduce another properties to AuthenticatedUser IMHO it should not be dump for everything that comes from configuration.

peterkvokacka-da (approved):

Issue #12, opened 2025-02-17
Canton components' histogram metrics are broken


vgangan:
All Canton **histogram** metrics that report `_bucket`s are absent except for `{le="+Inf"}` while at the same time `_sum` and `_count` suffixed siblings are present and contain correct values.

Canton config re: metrics overrides live [here](https://github.com/digital-asset/cn-quickstart/blob/c7c9a0e0396b8f03bf282b638b3eec798694e227/quickstart/compose.yaml#L14-L21)

Resolved config from the logs:
```
Starting up with resolved config:
...
monitoring {
    ...
    metrics {
        cardinality=20000
        histograms=[
            {
                aggregation {
                    max-buckets=160
                    max-scale=20
                    type=exponential
                }
                name="*"
            }
        ]
        jvm-metrics {
            enabled=yes
        }
        qualifiers=[
            errors,
            latency,
            saturation,
            traffic,
            debug
        ]
        reporters=[
            {
                address="0.0.0.0"
                port=14021
                type=prometheus
            }
        ]
    }
    ...
}
```

For example:
```
daml_executor_runtime_idle_duration_seconds_bucket{le!="+Inf"} <- series absent altogether
daml_executor_runtime_idle_duration_seconds_bucket{le="+Inf"} <- has data
daml_executor_runtime_idle_duration_seconds_sum <- has data
daml_executor_runtime_idle_duration_seconds_count <- has data
```

Prometheus querying:

<img width="1684" alt="Image" src="https://github.com/user-attachments/assets/bee2dd97-ae60-483d-8b07-9b5d5fe0d490" />

<img width="2117" alt="Image" src="https://github.com/user-attachments/assets/94e150e0-67c8-4a6a-894e-722a88e1a65f" />

<img width="2117" alt="Image" src="https://github.com/user-attachments/assets/ad7ebbe0-9cc1-44b9-905e-ac6566b3d497" />

<img width="2117" alt="Image" src="https://github.com/user-attachments/assets/bb9b93f9-70f1-4bdf-a3c0-c329498183e0" />

vgangan:
@hrischuk-da , assigning to you to liaise with correct members of the Canton/CN teams to get this resolved.

daravep:
heya, i've asked @nicu-da whether he can spot why it isn't working, as the CN team is using it. his comment is:
```
[...]

Issue #13, opened 2025-02-17
Some observability dashboards are broken due to missing Canton metrics


vgangan:
Currently the following observability dashboards & associated metrics are not rendering any data:
```
Participant:
- Caching
	+ daml_cache_misses
	+ daml_cache_evictions
- Commands
	+ daml_commands_failed_command_interpretations_total
- Dirty Requests
	+ daml_dirty_requests
	+ daml_max_dirty_requests
```

These metrics are absent altogether in the [CN release](https://github.com/digital-asset/cn-quickstart/blob/5fbbec19566b075276282dfbcdc9e35ebe22cb3c/quickstart/.env#L24) we consume:
```
IMAGE_TAG=0.3.3-snapshot.20241204.7793.0.veab38e18
```

We need someone from Canton side to review the metrics above and suggest if any of them have a new name or are gone altogether without replacement. We then need to decide if we need to remove any dashboards or author new ones that are relevant to the new reality.

vgangan:
@hrischuk-da , assigning to you to liaise with correct members of the Canton/CN teams to get this resolved (and make a call if any of the dashboards would need to get modified as a result).

vgangan:
@daravep I can't guess what the new metrics should be, for obvious reasons. Hence the request to engage someone knowledgeable from Canton's side to create the mapping for the problematic timeseries reported above:
```
foo -> bar
baz -> qux
qwe -> <gone, no match, suggestion - remove the panel>
```

vgangan:
OK, based on the docs edits in https://github.com/digital-asset/cn-quickstart/issues/13 I've been able to find the missing metrics. I have adjusted the list accordingly - left the ones I could not wrap my head around if they still exist. Would appreciate it if someone could let me know if any of these are still relevant.

daravep:
@vgangan cache metrics still exist but they are "debug" only. I would disregard. Most caches in canton are caffeine based which isn't wired up to the metrics, so the values you are seeing are useless.

the commands still exist, i think they are now "daml_api_commands_failed_command_interpretations_total"

dirty requests got renamed to 
daml_participant_inflight_validation_requests
daml_participant_max_inflight_validation_requests

PR #16, opened 2025-02-19
QS-12, QS-13: fix observability dashboards

vgangan:
This PR fixes somewhat broken dashboards in Quickstart. 2 main reasons for absent metrics:
* CN emits exponential histograms now instead of classic
* `participant-sv` node's was misconfigured in terms of metrics and not scraped by `otel-collector`

There is still an outstanding request to clarify 5 missing metrics (tracked by https://github.com/digital-asset/cn-quickstart/issues/13) so a small number of dashboards currently are empty, but otherwise the rest are present.

This PR effectively resolves https://github.com/digital-asset/cn-quickstart/issues/12

True source of dashboards code is https://github.com/DACH-NY/grafana-tools/pull/5, in this PR we just include the renders.


pgarmaz-da (approved):
LGTM

PR #25, opened 2025-02-19
Security and auth rework

peterkvokacka-da:
- Reworked authentication using `keycloak` that supports OAuth2/OIDC.
   - Replaced `mock oauth2 server` with `keycloak`
- Created `nginx-keycloak` to work around some issue with keycloak and to solve problem caused by usage of localhost subdomain `keycloak.localhost`
- Configured proper logout that not only remove session from backend-service but also follow OAauth2/OIDC spec to do `Front-Channel Logout` 
- Cleanup security in backend-service
   - Decoupled business code in backend-service from OAuth2 impl classes, coding to well defined interfaces instead
- Updated `TenantRegistration` fronted/backend
- Added support for backend-service admin endpoints to be called from CLI using JWT tokens
- Reworked onboarding/auto configuration of components
   - almost every auto configuration script runs in separate container e.g. `validator-app-user-auto-config` in preparation for https://github.com/digital-asset/cn-quickstart/issues/32 and to better demonstrate what environment variables are strictly necessary for configuration of containers needed in prod and what are just env variables needed for auto-config
- Removed all `bootstrap.sc` and need for `participant.init.auto-init = true` using JSON API V2 instead.
- Cleaned up backend-service AppProvider configuration
   - It starts now with bare minimum conf no pre-registered AppUser tenants. (New tenant can be registered dynamically later when AppUser node starts up calling TenantRegistration endpoint from CLI or web UI).

http://keycloak.localhost:8082/admin/master/console/#/master admin/admin
# app-provider
Bob participant/wallet admin

# app-user
Alice participant/wallet admin

password for all users: secret

Both accounts for validator clients:
service-account-app-provider-validator
service-account-app-user-validator
have right ParticipantAdmin

```
# app-provider
participant.ledger_api.users.list()
res0: UsersPage = UsersPage(
  users = Vector(
    User(
      id = "0145df12-c560-40fa-bef3-caff2c5c2224",
      primaryParty = None,
      isActive = true,
      annotations = Map("username" -> "service-account-app-provider-pqs"),
      identityProviderId = ""
    ),
    User(
      id = "1a36eb86-4ccc-4ec6-b7b7-caa08b354989",
      primaryParty = None,
      isActive = true,
      annotations = Map("username" -> "service-account-app-provider-backend"),
      identityProviderId = ""
    ),
    User(
      id = "553c6754-8879-41c9-ae80-b302f5af92c9",
      primaryParty = Some(value = app_provider_quickstart-peterkvokacka-1::1220b962c90c...),
      isActive = true,
      annotations = Map("username" -> "bob"),
      identityProviderId = ""
    ),
    User(
      id = "c87743ab-80e0-4b83-935a-4c0582226691",
      primaryParty = Some(value = app_provider_quickstart-peterkvokacka-1::1220b962c90c...),
      isActive = true,
      annotations = Map("username" -> "service-account-app-provider-validator"),
      identityProviderId = ""
    )
  ),
  nextPageToken = ""
)

# app-user
participant.ledger_api.users.list()
res0: UsersPage = UsersPage(
  users = Vector(
    User(
      id = "92a520cb-2f09-4e55-b465-d178c6cfe5e4",
      primaryParty = Some(value = app_user_quickstart-peterkvokacka-1::12207625eb8e...),
      isActive = true,
      annotations = Map("username" -> "alice"),
      identityProviderId = ""
    ),
    User(
      id = "97bb6cef-a7a9-410b-ba8c-ada08451a5c9",
      primaryParty = Some(value = app_user_quickstart-peterkvokacka-1::12207625eb8e...),
      isActive = true,
      annotations = Map("username" -> "service-account-app-user-validator"),
      identityProviderId = ""
    )
  ),
  nextPageToken = ""
)
```


quickstart/compose.yaml
```diff
@@ -317,17 +423,23 @@ services:
...
+      AUTH_APP_PROVIDER_TOKEN_URL: "${AUTH_APP_PROVIDER_TOKEN_URL}"
+      AUTH_APP_PROVIDER_PARTICIPANT_ADMIN_USER_NAME: "${AUTH_APP_PROVIDER_PARTICIPANT_ADMIN_USER_NAME}"
+      AUTH_APP_PROVIDER_PARTICIPANT_ADMIN_USER_PASSWORD: "${AUTH_APP_PROVIDER_PARTICIPANT_ADMIN_USER_PASSWORD}"
+      AUTH_APP_PROVIDER_VALIDATOR_USER_ID: "${AUTH_APP_PROVIDER_VALIDATOR_USER_ID}"
+      AUTH_APP_USER_BACKEND_OIDC_CLIENT_ID: "${AUTH_APP_USER_BACKEND_OIDC_CLIENT_ID}"
+      AUTH_APP_USER_ISSUER_URL: "${AUTH_APP_USER_ISSUER_URL}"
+      AUTH_APP_USER_TOKEN_URL: "${AUTH_APP_USER_TOKEN_URL}"
+      AUTH_APP_USER_PARTICIPANT_ADMIN_USER_NAME: "${AUTH_APP_USER_PARTICIPANT_ADMIN_USER_NAME}"
+      AUTH_APP_USER_PARTICIPANT_ADMIN_USER_PASSWORD: "${AUTH_APP_USER_PARTICIPANT_ADMIN_USER_PASSWORD}"
+      AUTH_APP_USER_VALIDATOR_USER_ID: "${AUTH_APP_USER_VALIDATOR_USER_ID}"
```
meiersi-da, line 441: Why are these here? these should be app-user-local concerns.

quickstart/compose.yaml
```diff
@@ -390,16 +505,21 @@ services:
           }
       - |
         ADDITIONAL_CONFIG_AUTO_INIT=
-        canton.participants.participant.init.auto-init=true
+        canton.participants.participant {
+          init.auto-init=true
```
meiersi-da, line 509: @peterkvokacka-da : why is this required?

AFAIS, this is not done in the splice compose files: https://github.com/DACH-NY/canton-network-node/blob/main/cluster/compose/validator/compose.yaml

and AFAIR, the validator app relies in some cases about the participant node not initializing on its own.

meiersi-da: @moritzkiefer-da: do you know what exactly these cases are? Is the validator app even compatible with an auto-inited participant?

meiersi-da: @peterkvokacka-da : a quick check on my machine showed that all pods come up health w/o auto-init, but the await script is blocked on some condition.

So it seems you should be able to remove the auto-init; and drop the canton bootstrap scripts, which is a good thing. Bootstrap scripts are notoriously hard to make reliable across partial bootstrapping.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
```diff
@@ -55,9 +55,9 @@ public CompletableFuture<ResponseEntity<AuthenticatedUser>> getAuthenticatedUser
         // Create the AuthenticatedUser object
         AuthenticatedUser user = new AuthenticatedUser(
                 // name
-                party.split("::")[0],
+                auth.getPrincipal().getAttribute("name"),
                 // party
-                party,
+                auth.getPrincipal().getAttribute("party"),
```
meiersi-da, line 60: How does this exactly work? How is that "party" attribute set and where does it come from?

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/OAuth2AuthenticationSuccessHandler.java
```diff
@@ -13,42 +13,53 @@
...
         ClientRegistration clientReg = authorizedClientService.loadAuthorizedClient(auth.getAuthorizedClientRegistrationId(), auth.getName()).getClientRegistration();
 
         List<GrantedAuthority> authorities = new ArrayList<>();
-        if (appProviderTokenUri.equals(clientReg.getProviderDetails().getTokenUri())) {
+        if (appProviderIssuerURL.equals(clientReg.getProviderDetails().getIssuerUri())) {
             authorities.add(new SimpleGrantedAuthority("ROLE_ADMIN"));
         }
 
+        Map<String, Object> claimsWithParty = new HashMap<>(oidcUser.getClaims());
+        claimsWithParty.put("party", clientReg.getClientName());
```
meiersi-da, line 57: This looks fishy, client names should not be party-ids as per our last discussion. Is this a temporary state? i.e., are you missing a TODO here?

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
meiersi-da: I now see it: it is put in as an extra claim by the `...AuthSuccessHandler` above: https://github.com/digital-asset/cn-quickstart/pull/25/files#diff-53db2ce42e2ac7234eabf3cb164afdecb988c65ced2a18989384879509093eddR57

Consider making this an `_virtual_claim_client_party_id` so it is easily grepable.

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/OAuth2AuthenticationSuccessHandler.java
meiersi-da: I guess quite a bit of this code here will change once you change the configuration format to configure an app-user's party-id together with all the other client-metadata correct?

quickstart/.env
```diff
@@ -90,9 +175,10 @@ SCRIBE_POSTGRES_HOST=${SPLICE_APP_PROVIDER_DB_SERVER}
 SCRIBE_POSTGRES_USER=${SPLICE_APP_PROVIDER_DB_USER}
 SCRIBE_POSTGRES_PASSWORD=${SPLICE_APP_PROVIDER_DB_PASSWORD}
 SCRIBE_POSTGRES_DATABASE=scribe
-SCRIBE_PIPELINE_OAUTH_CLIENTID=${AUTH_APP_PROVIDER_CLIENT_ID}
-SCRIBE_PIPELINE_OAUTH_CLIENTSECRET=${AUTH_APP_PROVIDER_CLIENT_SECRET}
-SCRIBE_PIPELINE_OAUTH_ENDPOINT=${AUTH_APP_PROVIDER_TOKEN_URI}
+SCRIBE_PIPELINE_OAUTH_CLIENTID=${AUTH_APP_PROVIDER_PQS_CLIENT_ID}
+SCRIBE_PIPELINE_OAUTH_CLIENTSECRET=${AUTH_APP_PROVIDER_PQS_CLIENT_SECRET}
+SCRIBE_PIPELINE_OAUTH_SCOPE=openid
+SCRIBE_PIPELINE_OAUTH_ENDPOINT=${AUTH_APP_PROVIDER_TOKEN_URL}
```
meiersi-da, line 181: Not on this PR, but on another one: consider using `SCRIBE_APP_PROVIDER` as the prefix to make it clear who runs this.

quickstart/.env
```diff
@@ -28,31 +28,116 @@ AUTH_URL=overriden_by_specific_variable
...
+AUTH_APP_PROVIDER_ISSUER_URL=http://keycloak.localhost:8082/realms/AppProvider                            # for backend, wallet-ui
+                                                                                                          # - needed for demo purposes only (we are using localhost subdomains)
+                                                                                                          # In production there will be a real addressable (also from backend) domain (or more)
+
+AUTH_APP_PROVIDER_WELLKNOWN_URL=${AUTH_APP_PROVIDER_ISSUER_URL_BACKEND}/.well-known/openid-configuration          # for validator
+AUTH_APP_PROVIDER_TOKEN_URL=${AUTH_APP_PROVIDER_ISSUER_URL_BACKEND}/protocol/openid-connect/token                 # for scribe, canton-console, onboarding
+AUTH_APP_PROVIDER_JWK_SET_URL=${AUTH_APP_PROVIDER_ISSUER_URL_BACKEND}/protocol/openid-connect/certs               # for participant, validator
+
+# app provider users
+AUTH_APP_PROVIDER_PARTICIPANT_ADMIN_USER_ID=553c6754-8879-41c9-ae80-b302f5af92c9                          # for participant, onboarding | has ParticipantAdmin right
```
meiersi-da, line 41: These hard-coded strings raise the question of where they come from, and how a dev using CN quickstart would change them, and how so? Or do they not need to change this?

Seems like a little intro comment at the top of this huge env file would be helpful.

Also consider splitting the env file into one per trust-domain: SV, app provider, app user

That should make it quite a bit easier to grok.

quickstart/.env
```diff
@@ -28,31 +28,116 @@ AUTH_URL=overriden_by_specific_variable
...
+APP_USER_VALIDATOR_PARTICIPANT_ADDRESS=participant-app-user
+
+LEDGER_API_AUTH_AUDIENCE=SV_https://canton.network.global
+LEDGER_API_ADMIN_USER=V_ledger-api-user
+VALIDATOR_AUTH_CLIENT_ID=V_ledger-api-user
+VALIDATOR_AUTH_CLIENT_SECRET=V_secret
+WALLET_ADMIN_USER=V_administrator
+VALIDATOR_AUTH_AUDIENCE=V_https://canton.network.global
+WALLET_UI_CLIENT_ID=V_wallet-ui-client-id
+ANS_UI_CLIENT_ID="V_"
```
meiersi-da, line 139: What are these variable used for? Why are they not scoped to one of the trust domains: SV, app user, app provider? Why is the "V_" prefix used in most of them, but the first one uses "SV_" ?

meiersi-da (commented):
Great work @peterkvokacka-da :clap: 

I've added a bunch of comments, but to my untrained eye this looks good. I suspect you'll discover a lot more about this once you attempt to do a secure production deployment of this, and write the guide for how to do so for your CN quickstart devs building on quickstart.

Also: I'd suggest that you @peterkvokacka-da ping @moritzkiefer-da tomorrow after you've addressed the comments that you can for a review. This way we should be in a good position to iron out the last set of questions on Monday in the meeting I've setup.

On

> ```
> # app-provider
> participant.ledger_api.users.list()
> res0: UsersPage = UsersPage(
>   users = Vector(
>     User(id = "0145df12-c560-40fa-bef3-caff2c5c2224", primaryParty = None, isActive = true, annotations = Map("username" -> "service-account-app-provider-pqs"), identityProviderId = ""),
>     User(id = "1a36eb86-4ccc-4ec6-b7b7-caa08b354989", primaryParty = None, isActive = true, annotations = Map("username" -> "service-account-app-provider-backend-to-participant"), identityProviderId = ""),
>     User(id = "3c467325-772f-4f93-8972-711ec75634b1", primaryParty = Some(value = quickstart-peterkvokacka-1::1220ee0e6aaf...), isActive = true, annotations = Map("username" -> "beatris"), identityProviderId = ""),
>     User(id = "553c6754-8879-41c9-ae80-b302f5af92c9", primaryParty = Some(value = quickstart-peterkvokacka-1::1220ee0e6aaf...), isActive = true, annotations = Map("username" -> "bob"), identityProviderId = ""),
>     User(id = "c87743ab-80e0-4b83-935a-4c0582226691", primaryParty = Some(value = quickstart-peterkvokacka-1::1220ee0e6aaf...), isActive = true, annotations = Map("username" -> "service-account-app-provider-validator"), identityProviderId = ""),
>     User(id = "participant_admin", primaryParty = None, isActive = true, annotations = Map(), identityProviderId = "")
>   ),
>   nextPageToken = ""
> )
> 
> # app-user
> participant.ledger_api.users.list()
> res0: UsersPage = UsersPage(
>   users = Vector(
>     User(id = "92a520cb-2f09-4e55-b465-d178c6cfe5e4", primaryParty = Some(value = quickstart-peterkvokacka-1::12203ddfaed1...), isActive = true, annotations = Map("username" -> "alice"), identityProviderId = ""),
>     User(id = "97bb6cef-a7a9-410b-ba8c-ada08451a5c9", primaryParty = Some(value = quickstart-peterkvokacka-1::12203ddfaed1...), isActive = true, annotations = Map("username" -> "service-account-app-user-validator"), identityProviderId = ""),
>     User(id = "e37a3590-77d7-4acf-a053-70e84211dfa6", primaryParty = Some(value = quickstart-peterkvokacka-1::12203ddfaed1...), isActive = true, annotations = Map("username" -> "adrian"), identityProviderId = ""),
>     User(id = "participant_admin", primaryParty = None, isActive = true, annotations = Map(), identityProviderId = "")
>   ),
>   nextPageToken = ""
> )
> ```

Consider deactivating the `participant_admin` users in favor of `bob` and `alice` as soon as you've finished bootstrapping bob and alice. Thereby avoiding "dead" users with potentially outdated credentials.

quickstart/docker/oauth/data/AppProvider-realm.json
```diff
@@ -0,0 +1,2350 @@
+{
```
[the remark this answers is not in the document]
peterkvokacka-da: These data files are exported from keycloak and imported on keycloak startup. That was the easiest solution to start playing with keycloak. So if they need to adjust for now I suggest to use keycloak web ui http://keycloak.localhost:8082/admin/master/console/#/master admin/admin 
and export data through command line  e.g. `/opt/keycloak/bin/kc.sh export --dir=/tmp/export --realm AppUser`
In production of course they will define keycloak DB and configuration there should survive container restart.

I am open to any suggestions if you have better idea how to do this. At the time I opted for this it was good enough solution and I had other priorities.

quickstart/backend/src/main/java/com/digitalasset/quickstart/ledger/ScanProxyApiConfiguration.java
```diff
@@ -13,7 +13,7 @@
 @Configuration
 public class ScanProxyApiConfiguration {
 
-    private static final String CLIENT_REGISTRATION_ID = "AppProvider-client-credentials";
+    private static final String CLIENT_REGISTRATION_ID = "AppProviderBackend-Participant"; // TODO change to validator when we have different audiences
```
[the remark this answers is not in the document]
peterkvokacka-da: taken from [docs](https://docs.dev.sync.global/validator_operator/validator_helm.html#oidc-provider-requirements)
```
OIDC_AUTHORITY_LEDGER_API_AUDIENCE | The audience for the participant ledger API. e.g. https://ledger_api.example.com

OIDC_AUTHORITY_VALIDATOR_AUDIENCE | The audience for the validator backend API. e.g. https://validator.example.com/api

```
`app-user` is already configure like that there are different audiences for participant and validator. `app-provider` should be adjusted like that as well.

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/Auth.java
```diff
@@ -0,0 +1,8 @@
+package com.digitalasset.quickstart.oauth;
+
+public final class Auth {
+  // TODO
```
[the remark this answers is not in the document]
peterkvokacka-da: This is still WIP. I have plan to consolidate Auth related stuff to ideally one place but since there are lot of changes already happening in this PR I would probably do it in future PR. This merely reminder for me to do that.

quickstart/.env
```diff
@@ -28,31 +28,116 @@ AUTH_URL=overriden_by_specific_variable
...
+AUTH_APP_PROVIDER_PQS_USER_NAME=service-account-app-provider-pqs
+AUTH_APP_PROVIDER_PQS_USER_ID=0145df12-c560-40fa-bef3-caff2c5c2224                                        # has ReadAs and ActAs rights as APP_PROVIDER_PARTY
+
+AUTH_APP_PROVIDER_ONBOARDING_CLIENT_ID=app-provider-unsafe                                                # for onboarding/demo purposes not suitable for production!
+                                                                                                          # Uses OAuth2 Resource Owner Password Credentials Grant Flow
+                                                                                                          # Used to automatically obtain information for onboarding
+                                                                                                          # that would be otherwise entered manually by an Administrator
+
+# app provider audiences
+# TODO set different audiences
```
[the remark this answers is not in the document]
peterkvokacka-da: this is just for me WIP will be done in this PR eventually. See comment about audiences above.

quickstart/compose.yaml
```diff
@@ -59,19 +60,96 @@ services:
...
+
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.1.0
+    container_name: keycloak
+    volumes:
+      - ./docker/oauth/data:/opt/keycloak/data/import
+    environment:
+      KEYCLOAK_ADMIN: admin
+      KEYCLOAK_ADMIN_PASSWORD: admin
+#      KC_DB: postgres
```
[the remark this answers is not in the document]
peterkvokacka-da: Oh I just tested whether KC can import realms from json files directly to H2 in memory DB. I would probably return it back as export KC that uses H2 has some issue. Export from KC with postgres works fine.

quickstart/compose.yaml
```diff
@@ -85,10 +163,26 @@ services:
...
+      - AUTH_APP_PROVIDER_BACKEND_TO_PARTICIPANT_USER_NAME=${AUTH_APP_PROVIDER_BACKEND_TO_PARTICIPANT_USER_NAME}
+      - AUTH_APP_PROVIDER_TOKEN_URL=${AUTH_APP_PROVIDER_TOKEN_URL}
+      - AUTH_APP_USER_PARTICIPANT_ADMIN_USER_ID=${AUTH_APP_USER_PARTICIPANT_ADMIN_USER_ID}
+      - AUTH_APP_USER_PARTICIPANT_ADMIN_USER_NAME=${AUTH_APP_USER_PARTICIPANT_ADMIN_USER_NAME}
+      - AUTH_APP_USER_PARTICIPANT_ADMIN_USER_PASSWORD=${AUTH_APP_USER_PARTICIPANT_ADMIN_USER_PASSWORD}
+      - AUTH_APP_USER_VALIDATOR_USER_ID=${AUTH_APP_USER_VALIDATOR_USER_ID}
+      - AUTH_APP_USER_WALLET_ADMIN_USER_NAME=${AUTH_APP_USER_WALLET_ADMIN_USER_NAME}
+      - AUTH_APP_USER_WALLET_ADMIN_USER_ID=${AUTH_APP_USER_WALLET_ADMIN_USER_ID}
+      - AUTH_APP_USER_WALLET_ADMIN_USER_PASSWORD=${AUTH_APP_USER_WALLET_ADMIN_USER_PASSWORD}
+      - AUTH_APP_USER_TOKEN_URL=${AUTH_APP_USER_TOKEN_URL}
```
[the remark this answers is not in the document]
peterkvokacka-da: Yes that is intended. there are just two onboarding containers one for participants and one for validators and everything else. Wanted to avoid starting many onboarding containers.

quickstart/compose.yaml
peterkvokacka-da: backend service needs for configuration parties for app-provider and app-user. Variables here are used in /backend-service/start.sh 

It can be simplified a bit now as I assigned primaryParties also to participantAdmins. Good point.

quickstart/compose.yaml
peterkvokacka-da: I didn't introduce that actually. It was there probably from time when we used bootstrap.sc. We don't use them anymore I didn't realise we can get rid of this. Thx.
I will test it and remove if not needed.

quickstart/.env
peterkvokacka-da: Those values are coming from keycloak configuration. If they want to use different OIDC or they change conf in keycloak they would need to adjust them. And I suggest to adjust them!!! There are passwords there as well.

I agree that comments and doc update needs more work.
Good idea to split the file.

quickstart/.env
peterkvokacka-da: WIP> Those are relicts from previous configuration. Still some of them are used in upstream compose.yaml files. 
Just wanted to see if something blows up using them. Will remove them or set to `overriden_by_specific_variable`

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/OAuth2AuthenticationSuccessHandler.java
peterkvokacka-da: @iggy was working in parallel on PR that introduces TenantProperties I tried to coordinate the work with the idea state above but it was hard. So I opted just to unblock Iggy with his PR and this will be addressed in future PR together with other code polishing when things are more stable with auth.

quickstart/compose.yaml
peterkvokacka-da: Just checked it is still needed otherwise participant will not start correctly. There is `bootstrap-entrypint.sc` in canton-participant.
```
import $file.bootstrap

logger.info("=== Bootstrapping application ===")

bootstrap.main()

logger.info("=== Bootstrapping complete ===")
```

moritzkiefer-da: You can initialize a validator with auto-init but you break hard domain migrations which is obviously not an option outside of simple tests. So I strongly recommend to nuke this.

quickstart/backend/src/main/java/com/digitalasset/quickstart/ledger/ScanProxyApiConfiguration.java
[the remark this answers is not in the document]
moritzkiefer-da: iirc the utility is not compatible with different audiences. So you may wanna default to using the same.

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/OAuth2AuthenticationSuccessHandler.java
meiersi-da: Sounds good. Adding a TODO in the code pointing this out would have made me not ask the question ;-)

quickstart/.env
```diff
@@ -11,8 +11,8 @@ DOCKER_NETWORK=quickstart
 # canton
```
moritzkiefer-da, line 11: probably not for this PR but this file seems to be getting out of hand. You probably wanna at a minimum split it, e.g., into env for app provider, end user, … and I'd also consider using some sort of yaml config instead where you can have a decent hierarchy

quickstart/docker/await-onboarding-done/participant-onboarding.sh
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: 0BSD
+
+set -eo pipefail
+
+source /app/utils.sh
+
+# Create admin user for validator oauth2 client as it is required by validator service configuration
+# As we are not connected to synchronizer yet (that will happen in validator initialization) we create user without primary party.
+# TODO comment just for reviewers remove before merge
+# Validator client needs ParticipantAdmin rights otherwise:
```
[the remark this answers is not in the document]
moritzkiefer-da: just to be clear: both of your examples go through the admin API where currently the validator doesn't support auth at all. The ParticipantAdmin right is still required for allocating parties and users though.

quickstart/.env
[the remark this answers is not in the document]
meiersi-da: ack. One trick that we use in CN to make it more clear what is meant to be addressed in the current PR and what is meant to be addressed in future changes is to use
- `FIXME` for current PR snags
- `TODO(#<issue-nr>)` for future changes

We also run CI checks that disallow FIXME's and enforce an open issue reference for every TODO. We copied and adjusted that from the `canton` repo.

Code here in case you want to do something similar: https://github.com/DACH-NY/canton-network-node/blob/main/.circleci/todo/src/checkTodos.sc

quickstart/docker/oauth/data/AppProvider-realm.json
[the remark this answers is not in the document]
meiersi-da: This sounds fine to me. 

So my question is more of a request for documentation. Inline code comments are a fine form of documentation, as would be a `README.md` file inside the `oauth/data` dir.

quickstart/compose.yaml
meiersi-da: OK, I see. 

This warrants a comment along the lines of:
```
Note that the `backend` config is used to specify the party-id associated with an app user,
which is why we have to pass it in here.
This won't be required once the state of an
onboarded app user is moved from the
backend configuration file into a backend accessible DB.
```

quickstart/compose.yaml
meiersi-da: @moritzkiefer-da : that sounds as if we should make sure to fail to initialize the validator app if the participant has been auto-initialized, provided we can detect this?

moritzkiefer-da: I think we have an issue for it somewhere it's a bit tricky to make work though iirc.

quickstart/docker/utils.sh
```diff
@@ -4,6 +4,211 @@
 
 set -eo pipefail
 
+get_app_provider_user_token() {
+  local user=$1
+  local password=$2
+  echo "get_app_provider_user_token $user" >&2
+
+  curl -f -s -S "${AUTH_APP_PROVIDER_TOKEN_URL}" \
```
moritzkiefer-da, line 12: This looks quite dodgy to me. It seems like you are trying to get a token for an end user which is usually only set up for an authorization code flow through a keycloak specific flow. That doesn't translate to auth0 or other IAMs.

If you need to get a token through curl I would expect that you set up a client credentials flow and use that which is standardized and works across IAMs.

Note that you cannot assume that you get the same subject for the client credentials flow that you get for the authorization code flow (auth0 doesn't allow you to do that at all for example). The way we handle this usually is to have two different ledger API users with the same parameters.

quickstart/compose.yaml
peterkvokacka-da: When I commented out 
```
#        ADDITIONAL_CONFIG_AUTO_INIT=
#          canton.participants.participant.init.auto-init=true
```
for `participant-app-provider`

I got problem in onboarding script
```
2025-02-21 19:24:52 get_app_provider_user_token bob
2025-02-21 19:24:54 create_user c87743ab-80e0-4b83-935a-4c0582226691 service-account-app-provider-validator  participant-app-provider
2025-02-21 19:24:54 http://participant-app-provider:7575/v2/users
2025-02-21 19:24:54 --data-raw {
2025-02-21 19:24:54       "user" : {
2025-02-21 19:24:54           "id" : "c87743ab-80e0-4b83-935a-4c0582226691",
2025-02-21 19:24:54           "isDeactivated": false,
2025-02-21 19:24:54           "primaryParty" : "",
2025-02-21 19:24:54           "identityProviderId": "",
2025-02-21 19:24:54           "metadata": {
2025-02-21 19:24:54              "resourceVersion": "",
2025-02-21 19:24:54               "annotations": {
2025-02-21 19:24:54                   "username" : "service-account-app-provider-validator"
2025-02-21 19:24:54               }
2025-02-21 19:24:54           }
2025-02-21 19:24:54       },
2025-02-21 19:24:54         "rights": [
2025-02-21 19:24:54         ]
2025-02-21 19:24:54     }
2025-02-21 19:24:54 curl: (7) Failed to connect to participant-app-provider port 7575 after 3 ms: Could not connect to server
```
even after many retries it didn't succeed.

quickstart/docker/utils.sh
peterkvokacka-da: This token is used only for automatic onboarding purpose. I do not expect anybody to use it in production this way that's why I named the special client created for this purpose `app-provider-unsafe`. In production I expect that whatever is configured using this token will be configured manually by admin.
image

moritzkiefer-da: I see. I'm a bit wary of that. My experience from CN and IAM in general is that it's very easy to bake in assumption that don't work out in prod. If this is really only for demo purposes this may be fine but then I think what I'm missing is validation that we can turn this into a prod setup (ideally not just tested against keycloak but also against auth0 which has different limitations) and clear docs for users on how to do so.

peterkvokacka-da: Look I am open to any suggestion that's why I asked you guys to review the PR in this early stage. Thank you for   the review! 

I understand your hesitation (even though this client really using standard OAuth2 flow) because I experienced myself how the OAuth2/OIDC spec is not very strict and is open for implementation interpretation more often than I like.

Take this as starting point. I needed to get to this point that I can demonstrate it works with keycloak somehow (and see the all the hurdles I needed to overcome) so we can start discussion otherwise it would very theoretical and impractical.

meiersi-da: @peterkvokacka-da : no worries! Understood that these are just first steps.

I believe the important question to align on is when to do what validation wrt a prod-level IAM setup.

@moritzkiefer-da comments convince me that this should be rather sooner than later. We don't want cn-quickstart users to get stuck on deploying to prod.

@hrischuk-da @filmackay : wdyt?

meiersi-da: > The CN QS does have the requirement to provide a working security model with JWT tokens and have documentation about how to configure the QS. The CN QS need to demonstrate that it can be run in TestNet and MainNet which are tasks yet to be done.

@hrischuk-da : sounds like we're broadly aligned. Avoiding blockers for CN quickstart customers on their road to MainNet seems important to do then to you as well, or not?

> I do think it would be helpful for @moritzkiefer-da and @meiersi-da to write out their learnings in the training or docs to help guide customers since the ETX team doesn't have the learned experience.

What we've learned is that systems like Auth0 have their own limitations; and that these limitations are surprising when just coming from a local deployment.

The answer to this lesson is to do a deployment of CN quickstart against Auth0 to see what is surprising for the app. We don't know all the surprises by heart, and likely the CN quickstart would hit different surprises than we hit for the CN core apps. 

We're advising for Auth0, as there we have evidence (from the 12+ SV node deployments in production) that a setup that works for Auth0 can be ported well to other IAMs (some SVs run on KeyCloak others run on AWS). 

If you are sure that customers will have a similarly simple porting experience for the KeyCloak-based setup used here, then that's fine as well.

quickstart/.env
```diff
@@ -28,31 +28,116 @@ AUTH_URL=overriden_by_specific_variable
...
+AUTH_APP_PROVIDER_VALIDATOR_USER_ID=c87743ab-80e0-4b83-935a-4c0582226691                                  # has ParticipantAdmin right
+
+AUTH_APP_PROVIDER_WALLET_UI_CLIENT_ID=app-provider-wallet                                                 # for wallet
+
+AUTH_APP_PROVIDER_PQS_CLIENT_ID=app-provider-pqs                                                          # for pqs/scribe
+AUTH_APP_PROVIDER_PQS_CLIENT_SECRET=zuYvMzWEo8csYNiQNlmXNPsmPErBWP3W
+AUTH_APP_PROVIDER_PQS_USER_NAME=service-account-app-provider-pqs
+AUTH_APP_PROVIDER_PQS_USER_ID=0145df12-c560-40fa-bef3-caff2c5c2224                                        # has ReadAs and ActAs rights as APP_PROVIDER_PARTY
+
+AUTH_APP_PROVIDER_ONBOARDING_CLIENT_ID=app-provider-unsafe                                                # for onboarding/demo purposes not suitable for production!
```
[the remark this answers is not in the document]
meiersi-da: Let's avoid just duplicating that information. Happy for docs to move, but let's figure out how me make sure that they are updated as changes flow in.

quickstart/docker/utils.sh
hrischuk-da: @meiersi-da  and I discussed in our 1-1.  TL;DR:
- I acknowledge that Auth0 integration is a risk and I'll be monitoring customers for signal that more is needed in the CN QS.
- We both agree there are higher priority items than this.
- The current authorization approach is sufficient and we should proceed.

quickstart/backend/src/main/java/com/digitalasset/quickstart/oauth/OAuth2AuthenticationSuccessHandler.java
peterkvokacka-da: reworked

quickstart/.env
peterkvokacka-da: removed

quickstart/.env
filmackay: It would also be worth referring to `SCRIBE` as `PQS` as this is the name we are focusing on 👍🏻

meiersi-da (approved):
Good work @peterkvokacka-da . Thanks for doing the refactoring wrt `TenantPropertiesRepository`; looks much better now.

The code looks sensible, but I'm wary whether the authentication and authorization rules work as expected w/o tests.

I'd suggest to either add such tests as part of this PR; or a follow-up PR to ensure the mechanisms work as expected. wdyt?

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/SecurityConfig.java
```diff
@@ -62,11 +80,30 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
...
+        JwtAuthenticationConverter converter = new JwtAuthenticationConverter();
+        converter.setJwtGrantedAuthoritiesConverter(new Converter<>() {
+            private final JwtGrantedAuthoritiesConverter defaultGrantedAuthoritiesConverter = new JwtGrantedAuthoritiesConverter();
 
+            @Override
+            public Collection<GrantedAuthority> convert(Jwt jwt) {
+                Collection<GrantedAuthority> authorities = new HashSet<>(defaultGrantedAuthoritiesConverter.convert(jwt));
+                // there is only one AppProvider issuer that can issue JWT to authenticate to ResourceServer
+                // we consider anybody with JWT from that issuer to be admin
+                authorities.add(new SimpleGrantedAuthority("ROLE_ADMIN"));
```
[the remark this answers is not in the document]
hrischuk-da: @meiersi-da Are you asking for an automated test or just proof that it works as expected (e.g., Peter manually hacks the QS to prove it works as expected)?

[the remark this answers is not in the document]
meiersi-da: I'm asking for automated tests.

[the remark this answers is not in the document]
hrischuk-da: Ok. Just to make sure we are aligned on scope, your ask is:
1. An automated test to show it enables authorized access (works in a positive case)
2. An automated test to show it prevents unauthorized access (the negative case)

filmackay (approved):

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/SecurityConfig.java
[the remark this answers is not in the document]
meiersi-da: > I obviously tested it manually and I think that is enough for our purpose. The test itself will not demonstrate how to use and test our technology but actually it will test spring-boot-starter-oauth2-resource-server and whether our [configuration of the Resource Server ](https://github.com/digital-asset/cn-quickstart/blob/a520b81926edceae06e0e571b5f7409dc0f70f2b/quickstart/backend/src/main/resources/application.yml#L4-L9)is correct.

Thanks for testing this manually. If it was my code, then I would have added the one automated test that shows that the configuration was indeed done correctly; and will remain correct going forward.

That said: this is not my code, so feel free to ignore my advice, which is geared towards optimizing dev velocity by having just enough tests to allow any dev to confidently change the code while just relying on CI to catch regressions.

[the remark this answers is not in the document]
peterkvokacka-da: It is very unlikely we see regression here. Somebody would need to mess with security configuration. First we need to cover with tests areas that regressing is quite possible. When we have testing infrastructure in place I may add tests you are asking for but atm it is not priority. Until then I'll keep an eye on security configuration in future PRs.

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/OAuth2ClientRegistrationRepository.java
```diff
@@ -0,0 +1,101 @@
...
+                        .redirectUri("{baseUrl}/login/oauth2/code/{registrationId}")
+                        .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
+                        .build()
+        );
+        return registrationId;
+    }
+
+    @Override
+    public void removeClientRegistration(String tenantId, String clientId) {
+        registrations.remove(tenantId + "-" + tenantId);
```
iggy-da, line 66: Not using clientId

quickstart/frontend/src/views/LoginView.tsx
```diff
@@ -37,6 +37,9 @@ const LoginView: React.FC = () => {
                 ))}
                 </tbody>
             </table>
+            {/*TODO get username/password from .env*/}
+            <div>AppProvider user: bob, password: secret</div>
+            <div>AppUser user: alice, password: secret</div>
```
iggy-da, line 42: Great!

Side-notes: In my opinion I don't think we have to `get username/password from .env` since this is not production code, it's fine like this. Maybe instead the comment should say something like `{/* Default credentials for convenience - remove for production use */}` and there could also be a small user-facing heading `Hint: Default Credentials` to give some context. Just a `nit`, does not have to be changed in this PR...

quickstart/docker/oauth/data/README.md
```diff
@@ -0,0 +1,68 @@
+# Keycloak export
```
iggy-da, line 1: Maybe it would be easier to find this file under the `docs` directory, and the document could then with time document other auth related details.

quickstart/docker/oauth/data/AppProvider-users-0.json
```diff
@@ -0,0 +1,92 @@
+{
```
iggy-da, line 1: Nit: The docker/oauth/data files could go into the `config` directory

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/UserApiImpl.java
```diff
@@ -118,6 +97,8 @@ public CompletableFuture<ResponseEntity<AuthenticatedUser>> getAuthenticatedUser
                 .exceptionally(ex -> {
                     Throwable cause = ex.getCause() != null ? ex.getCause() : ex;
                     if (cause instanceof SecurityException) {
+                        methodSpan.addEvent("User not authenticated");
+                        LoggingSpanHelper.logInfo(logger, "User is not authenticated");
```
iggy-da, line 101: :+1:

iggy-da (approved):
Great work!

quickstart/frontend/src/views/LoginView.tsx
iggy-da: That wouldn't make sense - let me explain. The backdrop is that Canton Network is tremendously complex and there is very little (no!) budget to spend on making `quickstart` any harder to get started with due to additional complexity. We need to aide the user through the process as much as possible.

What we are telling the user here is what to put into the boxes to get started. The alternative is to stump the user with blank logins fields, and initially no idea what to put into them, having to search the documentation, etc. This is usually easier said than done. People get confused easier than we imagine.

Yes, you would have this info in production, it's a trade-off decision. But in my opinion it is the sane thing to do if we want people to succeed with quickstart.

The reason why you would not want this to reflect the actual variable content is because doing so would pose a security risk. Logging the default username password is no risk at all, because this is publicly available information. But you don't want to log out the new password if you change it.

PR #28, opened 2025-02-21
HOTFIX:construct PARTY_HINT that contains only allowed characters

peterkvokacka-da:


iggy-da:
Thanks for finding this bug @peterkvokacka-da!
Addressed by #29 #30

PR #39, opened 2025-02-25
Update .github/CODEOWNERS to use ETX group

pgarmaz-da:
- update .github/CODEOWNERS to use @digital-asset/etx team instead of individual members
- add @digital-asset/etx to docs/ and README.md ownership


iggy-da (approved):
Thanks!

Issue #42, opened 2025-02-26
Add Swagger UI docker service to export JSON API V2 OpenAPI docs


judywu-da:
Add good to have feature to view participant API V2 docs.

PR #44, opened 2025-02-26
Add swagger-ui into docker network to allow viewing Canton JSON API V2 Open API spec in browser

judywu-da:
Purpose as titled
The PR is aimed to address the issue proposed [here](https://github.com/digital-asset/cn-quickstart/issues/42)

### DIFF
--- a/quickstart/Makefile
+++ b/quickstart/Makefile
@@ -181,6 +181,17 @@ install-daml-sdk: ## Install the Daml SDK
 generate-NOTICES: ## Generate the NOTICES.txt file
 	./gradlew generateNotices
 
+.PHONY: start-swagger-ui
+start-swagger-ui:
+	docker compose -f docker/swagger/compose.yaml $(DOCKER_COMPOSE_ENVFILE) up -d || true
+
+.PHONY: stop-swagger-ui
+stop-swagger-ui:
+	docker compose -f docker/swagger/compose.yaml $(DOCKER_COMPOSE_ENVFILE) down || true
+
+.PHONY: restart-swagger-ui
+restart-swagger-ui: stop-swagger-ui start-swagger-ui
+
 # Help target
 .PHONY: help
 help: ## Show this help message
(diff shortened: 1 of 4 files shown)


quickstart/config/nginx/app-provider/swagger.conf
```diff
@@ -0,0 +1,10 @@
+server {
+    listen 80;
+    server_name swagger.localhost;
+    location /docs/openapi {
+        proxy_pass http://participant-app-provider:7575/docs/openapi;
+        add_header Access-Control-Allow-Origin *;
```
judywu-da, line 6: Added headers here to allow CORS, avoiding the issue mentioned [here](https://github.com/swagger-api/swagger-ui/issues/8469)

quickstart/docker/swagger/compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+networks:
+  quickstart:
+    external: true
+    name: ${DOCKER_NETWORK}
+
+services:
+  swagger-ui:
+    image: swaggerapi/swagger-ui
+    ports:
+      - "9080:8080"
```
judywu-da, line 16: port to 9080 as 8080 is used for `backend service` already

judywu-da: not sure if it's necessary to make this port configurable in .env file 🤔 , lemme know if you think that's better. 🙏

quickstart/Makefile
```diff
@@ -181,6 +181,17 @@ install-daml-sdk: ## Install the Daml SDK
 generate-NOTICES: ## Generate the NOTICES.txt file
 	./gradlew generateNotices
 
+.PHONY: start-swagger-ui
+start-swagger-ui:
+	docker compose -f docker/swagger/compose.yaml $(DOCKER_COMPOSE_ENVFILE) up -d || true
```
iggy-da, line 186: I suggest to remove the `-d` flag and start the service in the foreground rather than having `start/stop/restart/clean` lifecycle targets (I would remove the other ones). It can then be terminated conventionally with `ctrl-c`. I think this is enough because the UI doesn't interact directly with the application as a service.

In that case it would make sense to log/`@echo` out the URL to open the UI before startup from the `make` target.

If instead you prefer the  existing lifecycle targets, please also add a `clean`-target and depend on it from `clean-docker`.

In either case please also add the `##` docstrings so that the targets show up in `make help` :pray:

judywu-da: Thanks Iggy, address the comment as suggested in [9ea95d3](https://github.com/digital-asset/cn-quickstart/pull/44/commits/9ea95d3cb73618bd34dfed7d449eb94ca759934a)

Issue #43, opened 2025-02-26
Add docker start script for starting docker desktop


judywu-da:
A convenient [script from canton-network-utility](https://github.com/DACH-NY/canton-network-utilities/blob/main/scripts/start-docker-desktop.sh) that start docker desktop before running any docker commands.

filmackay:
We will address this outcome/concern by detecting whether Docker is available and if not display an error to the effect of:

> Docker does not appear to be available, please run it as per https://docs.docker.com/engine/daemon/start/

This way we avoid the environmental concerns that would lead to specific solutions (eg. Docker Desktop) not working in various environments, as the operating environment is not homogeneous.

iggy-da:
See also #19

PR #46, opened 2025-02-26
Add docker start script for detecting/starting Docker service

judywu-da:
Add docker start script for detecting/starting Docker service.
The PR is aimed to address this [git issue](https://github.com/digital-asset/cn-quickstart/issues/43)

### DIFF
--- a/quickstart/scripts/start-docker-desktop.sh
+++ b/quickstart/scripts/start-docker-desktop.sh
@@ -15,8 +15,27 @@ echo "Checking if Docker is running..."
 
 # Check if Docker is running
 if ! is_docker_running; then
+  OS_TYPE="$(uname)"
   echo "Docker is not running. Starting Docker Desktop..."
-  open -a Docker
+  case "$OS_TYPE" in
+  
+  Darwin)
+    # macOS
+    open -a Docker;;
+
+  Linux)
+    # Linux
+    sudo systemctl start docker;;
+
+  MINGW*|CYGWIN*|MSYS*)
+    # Windows
+    powershell.exe -Command "Start-Service docker";;
+  
+  *)
+    echo "Unsupported operating system detected: $OS_TYPE"
+    echo "Please start Docker manually on your system."
+    exit 1;;
+  esac
 
   # Wait until Docker is running or stop after 5 iterations
   max_attempts=5
@@ -27,6 +46,7 @@ if ! is_docker_running; then
     attempt=$((attempt + 1))
     if [ $attempt -ge $max_attempts ]; then
       echo "Error: Docker did not start within the expected time after $max_attempts attempts."
+      echo "Please start Docker manually on your system."
       exit 1
     fi
   done
--- a/quickstart/scripts/start-docker-desktop.sh
+++ b/quickstart/scripts/start-docker-desktop.sh
@@ -2,7 +2,6 @@
 
 set -eou pipefail
 
-# Function to check if Docker is running
 is_docker_running() {
   if ! timeout 5 docker info > /dev/null 2>&1; then
     echo "Docker is not responding within the timeout period."
--- a/quickstart/scripts/start-docker-desktop.sh
+++ b/quickstart/scripts/start-docker-desktop.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+
+# Function to check if Docker is running
+is_docker_running() {
+  if ! timeout 5 docker info > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


quickstart/scripts/start-docker-desktop.sh
```diff
@@ -0,0 +1,35 @@
...
+  fi
+  return 0
+}
+
+echo "Checking if Docker is running..."
+
+# Check if Docker is running
+if ! is_docker_running; then
+  echo "Docker is not running. Starting Docker Desktop..."
+  open -a Docker
```
amarqueslee-da, line 19: This looks specific to desktop app deployments (likely, the Mac users), where the docker daemon is implicitly launched via the desktop app?

What if docker is installed with a different setup (for example, for a Linux user)?

Rather than taking ownership of the "how to run docker" problem, would it be simpler to de-scope this script to just tell the user that docker doesn't seem to be running and error-out with a hint?

judywu-da: Yeah, that's a good point, the script might not work on some Linux OS.


Note that this script is from [canton-network-utility](https://github.com/DACH-NY/canton-network-utilities/blob/main/scripts/start-docker-desktop.sh)
I quite like it when running the utility app, thus propose to do it here. 
Maybe we can try and add:
1. command to start it in Linux
2. some more protection in case the "start does not work on specific OS".
"Try starting by running <command> failed, please start Docker service manually."

filmackay: I think this is covered here: https://docs.docker.com/engine/daemon/start/

```
$ sudo systemctl start docker
```

Looks like there is no unified way to start it - needs an OS check :(

judywu-da: updated the script to support Linux and Windows to start docker service.
I have only tested it on my MacOS, wondering if people can help testing on other OS?
Note that for Windows, it won't work if it's not ran in a Unix-like subsystem like Git-bash, WSL.

amarqueslee-da: I believe @iggy-da uses Ubuntu.

meiersi-da:
@filmackay or @andrae-da given that @judywu-da is on holidays for the next 2 weeks, would one of you be able to help shepherd this valuable functionality into `main`? :pray:

iggy-da:
Thanks, this PR is done in good faith, by intuition is that we should not go down this route.

Managing your local docker service across all operating systems and environments is a tall order, if not impossible, and not `quickstart`'s responsibility or purpose. It makes sense to give the user meaningful error messages if docker is not available, and to link to relevant documentation. I think we should not go further than that.

To illustrate, I don't want open-source repos to run `sudo systemctl` commands on my machine, especially without asking. To such a repo I would say: You don't know my setup. You don't know if I'm using docker, podman, containerd, or whatever. You don't know if I'm running `systemd` or another service manager. You don't know the side-effects of running this `sudo` command on my machine.

There are so many variables that play into the the "right way to start the docker service". Not to mention that the script itself will not conventionally run on Windows, since it's a `bash` script.

This approach may work for company-internal repos, but I don't think it's a good fit for `quickstart`.

filmackay:
As an alternative approach, I think something with less risk would be to:

* detect if docker is running/responding
* display an error to the effect of:
> Docker does not appear to be available, please run it as per https://docs.docker.com/engine/daemon/start/

This has the advantage of:
* not getting involved in OS differences
* not getting involved in container deltas (docker/podman)
* not providing a script that needs `sudo` access (should be avoided?)
* not needing to debug specific environmental differences, which our experience says are many in corporate networks
* not having to follow changes in a range of OS's, different Docker distributions that will cause significant effort to resolve as DA support needs to investigate

judywu-da:
@filmackay should this PR be closed?

Or would you like to use it for the change mentioned above:
- detect if docker is running/responding
- display an error to the effect of:
> Docker does not appear to be available, please run it as per https://docs.docker.com/engine/daemon/start/

Also, saw that you un-assign me from this issue, does that mean etx team will prioritise and continue with the change, or would you like me to follow up with the changes?

Thanks 🙏

iggy-da:
@judywu-da
> * display an error to the effect of: Docker does not appear to be available, please run it as per https://docs.docker.com/engine/daemon/start/

Not Fil, but you are welcome to implement this if it helps address your original purpose for this PR or similar reasons (in that case, please reassign yourself). If not, please close this PR, and we can deal with this when it gets prioritized. Thank you! :pray:

judywu-da:
@iggy-da thanks for the reply!
We don't need it for now, will close the PR.

PR #54, opened 2025-03-05
Refactoring to unify UI and API

iggy-da:
The last PR in the series to address outstanding tasks addressing the code-quality of the existing code-base as prior work for #20 .

Addresses the feedback that both the UI and the API should target the domain-driven use-cases and user-needs rather than being build around the `daml` model. The four tabs related to the licensing model have been reduce to two, `AppInstalls` and `Licenses`, and the existing code has been merged and unified accordingly. The API went through similar changes. Notably, the API now indicates if `LicenseRenewalRequests` are paid, with an `isPaid` flag, which allows the UI to render `Pay` buttons on the user-side, and `Renew` buttons on the admin side, conditionally. The code now also uses Postgres `WHERE` clauses and `JOINS` instead of application-level filtering, where possible. The code has been judiciously restructured from big method monoliths into smaller, atomic methods to improve readability, where appropriate. The PR also includes other minor bug-fixes and quality-of-life improvements.


filmackay (approved):

peterkvokacka-da (approved):
Great. UI and backend code looks much cleaner. Thx.
/nit I would like to see action buttons without need to click on intermediary `Actions` button to discover what actions are available to me.

PR #66, opened 2025-03-13
adjust default users

peterkvokacka-da:
Curtis: I see "Alice Doe" and " Bob Doe".  I was wondering if we could change this to "Alice the user" and "Bob the provider". Also, can we change the password from 'secret' to something else like 'abc123'
Simon: using Pat the provider makes it easier to remember who's the provider


iggy-da (approved):
Thanks!

PR #77, opened 2025-03-18
PoC of colocating canton/splice components in a single JVM

peterkvokacka-da:
https://docs.google.com/document/d/1U_PYXoWHJ3QLN4s-wgr1LfPP4cTOJa0aB4vM6gfyI7E/edit?tab=t.0
This is first step towards the goals from the document verifying what that can look like in QS. Next step would be to distil Splice LocalNet out of QS.

TLDR: 
- canton components merged to one canton container
- splice components merged to one splice container
- postgres DBs also consolidated to one postgres container
- splice nginx containers merged to container

SV is optional (and turned off in devnet). AppUser, AppProvider could be made optional following the same approach if we want to. 

Users can adjust it as they need modifying the conf for canton, splice components.

Memory requirement down from 16gb to 6.3gb (without observability)
- bumped splice version to 0.3.18


iggy-da:
Just responding to the review request - is this a draft PR or ready for review? The relationship between the PR description and branch name/code is not obvious to me.

peterkvokacka-da:
sorry that suppose to be for another PR https://github.com/digital-asset/cn-quickstart/pull/78
This one is still wip

hrischuk-da:
@peterkvokacka-da I assume that is work that is going on in a branch so that a user can continue to use the QS.  Is that true?

filmackay:
> I'm still concerned with the disconnect from Splice configurations, which I assume will be addressed soon when you move this to Splice?

I think it would be good to at least agree on our intended approach - our goal here is that we get a simple LocalNet ("be like ethereum") that is easy to spin up and understand. Right now we are abusing an existing container.

At the end of the day we need a container to run LocalNet from, being one that could:

- be configured to operate, in a minimal configuration but containing all the essential defaults
- includes a/the library of HOCON components (eg. validator) that you already have (maintenance is the risk here though)

In this way we (LocalNet - and anyone wanting a flexible container) could inject a config that composes the library already available in-situ.

Alternatively, how would you like to close the gap with Splice config? We just found that the env-var based substitution was very brittle, and that HOCON far more flexible (mandatory?) when doing "everything in a JVM" because the usual layers of separation are not there.

Thoughts, @isegall-da ?

quickstart/docker/app-user-shell/scripts/simulate-user-input.sh
```diff
@@ -11,8 +11,8 @@ source /app/utils.sh
...
 
-APP_PROVIDER_PARTY=$(get_user_party "$APP_PROVIDER_PARTICIPANT_ADMIN_TOKEN" $AUTH_APP_PROVIDER_VALIDATOR_USER_ID participant-app-provider)
-APP_USER_PARTY=$(get_user_party "$APP_USER_PARTICIPANT_ADMIN_TOKEN" $AUTH_APP_USER_VALIDATOR_USER_ID participant-app-user)
+APP_PROVIDER_PARTY=$(get_user_party "$APP_PROVIDER_PARTICIPANT_ADMIN_TOKEN" $AUTH_APP_PROVIDER_VALIDATOR_USER_ID "canton:37575")
+APP_USER_PARTY=$(get_user_party "$APP_USER_PARTICIPANT_ADMIN_TOKEN" $AUTH_APP_USER_VALIDATOR_USER_ID "canton:27575")
 
 APP_USER_WALLET_ADMIN_TOKEN=$(get_user_token $AUTH_APP_USER_WALLET_ADMIN_USER_NAME $AUTH_APP_USER_WALLET_ADMIN_USER_PASSWORD $AUTH_APP_USER_AUTO_CONFIG_CLIENT_ID $AUTH_APP_USER_TOKEN_URL)
-DSO_PARTY=$(get_dso_party_id "$APP_USER_WALLET_ADMIN_TOKEN" validator-app-user)
\ No newline at end of file
+DSO_PARTY=$(get_dso_party_id "$APP_USER_WALLET_ADMIN_TOKEN" "splice:25003")
```
iggy-da, line 18: Nit: I think we should use interpolated variables where possible (`3${PARTICIPANT_JSON_API_PORT}`), otherwise the variables are useless because changing them does not get propagated, and it's worse than hard-coded values in some respects, because now you have to deal with both variants when changing the port...

I won't point out the other instances in bash files - if desired to clean this up, it's a job for search-and-replace tooling

quickstart/config/nginx/app-provider.conf
```diff
@@ -0,0 +1,26 @@
...
+  location / {
+    proxy_pass http://wallet-web-ui-app-provider:8080/;
+  }
+}
+
+server {
+    listen 3000;
+    server_name swagger.localhost;
+    location /docs/openapi {
+        proxy_pass http://canton:37575/docs/openapi;
```
iggy-da, line 21: Nit: Apparently the nginx container supports variable substitution (`$PARTICIPANT_JSON_API_PORT`) via envsubst in the config if you mount the file with a `.template` extension under `/etc/nginx/templates`. You may have to escape the existing `$` literals above

There are many other such instances in the `nginx` files. I won't point them out individually, since the replacement can be done with a search-and-replace in the `./config/nginx` directory for each port, if desired

README.md
```diff
@@ -115,7 +115,7 @@ Run `make help` to see a list of all available targets, including (but not limit
 
 ## Topology
 
-This diagram summarizes the relationship of services that are started as part of `make start`. The focus of `Canton Network Quickstart` is to provide a development environment for App Providers.
+This diagram summarizes the relationship of services that are started as part of `make start`. The two core services `canton` and `splice` contain multiple logical components inside. The `postgres` service contains multiple databases required by QS services. One `nginx` service is used as proxy for all QS services that needs one except for `keycloak` that has its own `nginx-keycloak` as it needs to be ready before other services start. The focus of `Canton Network Quickstart` is to provide a development environment for App Providers.
```
iggy-da, line 118:
```suggestion
This diagram summarizes the relationship of services that are started as part of `make start`. The `canton` and `splice` services are configured to serve multiple logically separate components (each component represented with a box in the diagram) from a single container to reduce resource consumption. Similarly the `postgres` service contains multiple databases required by QS services. One `nginx` service is used as proxy for all QS services that needs one except for `keycloak` that has its own `nginx-keycloak` as it needs to be ready before other services start. The focus of `Canton Network Quickstart` is to provide a development environment for App Providers.
```

Just a suggestion in an attempt to clarify further - feel free to use/adapt/reject.

iggy-da (approved):
Great work! :clap:  Thank you.

Side-note: I think the `config` and `docker` directories have lost all meaning and distinction. Originally, all configuration files were meant to go into `./config`. At some point the line started to blur, and by now they are interchangeable. It's out-of-scope for this PR, but I think we should create a ticket

filmackay:
> @peterkvokacka-da I assume that is work that is going on in a branch so that a user can continue to use the QS. Is that true?

@hrischuk-da once it merges, it goes to `main` and is no longer "on a branch". Users however can continue to use old versions of QS, for as long as the underlying dependency versions are still valid.

Finding an 'old version' involves them looking through commit history, unless we want to 'tag' a given commit and give it a human readable name. Not sure what we'd call it though as I'm not clear what concept we are trying to capture?

quickstart/compose.yaml
```diff
@@ -99,194 +116,165 @@ services:
...
-    environment:
-      - POSTGRES_USER=${SPLICE_APP_USER_DB_USER}
-      - POSTGRES_PASSWORD=${SPLICE_APP_USER_DB_PASSWORD}
+      - ./config/nginx/nginx.conf:/etc/nginx/nginx.conf
+      - ./frontend/dist/:/usr/share/nginx/html
+      - ./config/nginx/frontend.conf:/etc/nginx/conf.d/default.conf
+      - ./config/nginx/common-backend-proxy-settings.conf:/etc/nginx/common-backend-proxy-settings.conf
+      - ./config/nginx/app-provider.conf:/etc/nginx/conf.d/app-provider.conf
+      - ./config/nginx/app-user.conf:/etc/nginx/conf.d/app-user.conf
+      - ${SV_NGINX_CONFIG:-./docker/empty}:/etc/nginx/conf.d/sv.conf
```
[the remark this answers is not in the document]
peterkvokacka-da: The suggestion doesn't work well for me 
`invalid empty volume spec`

Issue #75, opened 2025-03-17
Auth improvements based on feedback from users


peterkvokacka-da:
Add curl functions:
- allocate_party
- onboard_wallet_user

Add env variable for keycloak import data dir so settings could be imported from different than default location.

PR #78, opened 2025-03-18
Small QoL improvements

peterkvokacka-da:
address #75


iggy-da:
Looks good, but what are these (so far) unused functions for? Not sure what I'm looking at, if you could help me with a sentence or so of context, that would be helpful.

peterkvokacka-da:
Those functions are needed if you need to onboard user to wallet (expect for wallet admin which is onboarded by default)

iggy-da (approved):
Thanks!

Assuming that the code is meant as an example snippet for users, consider moving it into a markdown file under `docs` with a short description, or if you think the code is best placed in `quickstart/docker/utils.sh`, add a comment inside the code. Otherwise it could be mistaken for "dead code".

PR #80, opened 2025-03-19
End to end integration testing

iggy-da:
The integration tests exercise the entire application in a fully containerized environment, verifying that all components (frontend, backend, Daml model, and various supporting services) work together as expected. These tests include UI interactions, API calls, and workflow validation against real services.

To run as a developer, run `make integration-test`. It won't interfere with the hosts application state (docker volumes/containers etc)

`integration-test/setup-and-run-tests.js` automates the management of an isolated Docker-in-Docker test runner, transferring proprietary images into that container, and executing the Playwright tests in a nested container.

For detailed information, please refer to the [testing readme](https://github.com/digital-asset/cn-quickstart/blob/end-to-end-integration-testing/docs/development/002-testing.md).

Implements #20


quickstart/common/openapi.yaml
```diff
@@ -170,36 +170,6 @@ paths:
         '400':
           description: Invalid input
 
-  /app-install-requests/{contractId}/cancel:
```
iggy-da, line 173: This didn't make sense since only the AppUser can cancel, and the backend doesn't have the right to act on their behalf

quickstart/Makefile
```diff
@@ -147,6 +147,22 @@ first-run-setup:
...
+.PHONY: integration-test
+integration-test: build ## Run integration tests
+	node ./integration-test/setup-and-run-tests.js
+
+.PHONY: integration-test-ci
+integration-test-ci: start ## Run integration tests using host's application deployment
+	@# The integration test expects four AppInstallRequests
+	for i in {1..4}; do \
+		make create-app-install-request; \
+	done
```
peterkvokacka-da, line 159: It works for current Licensing Workflow but can we include these manual steps into test suite itself and not have them as prerequisite for the test suite?
We may have in future manual steps in the middle workflow (e.g. AppUser may need to do manual steps for which UI doesn't exist in AppUser side other then AppInstallRequest)

iggy-da: > can we include these manual steps into test suite itself and not have them as prerequisite

Not possible at this stage without significant rework. The test-runner does not have access to the development environment, nor does it have the required binaries for the `make` commands. The test runner runs against the application itself, i.e. what's available over the network.

Since there is no API to act on the AppUser's behalf, the test-runner cannot perform this step unless we provide a networking service to do so, or redesign the runner setup so that it runs in the development environment. The latter is easier said than done: Among other things, this means that users have to install and configure their own browser installations instead of using the predictable docker image.

We can however remove the duplication of this setup between `integration-test/setup-and-run-tests.js` and this make target.

> AppUser may need to do manual steps for which UI doesn't exist in AppUser side other then AppInstallRequest

Agreed that's a good consideration. I suggest when, and if, such a case arises, we let that inform the approach we take to resolve this. I expect significant development cost and potentially reduced development experience for the end-user if preparing for this preemtively, for reasons touched on above.

docs/development/002-testing.md
```diff
@@ -0,0 +1,140 @@
...
+
+The integration tests exercise the entire application in a **fully containerized environment**, verifying that all components (frontend, backend, Daml model, and various supporting services) work together as expected. These tests include UI interactions, API calls, and workflow validation against real services.
+
+Two main approaches are supported, depending on your use case:
+
+1. **`make integration-test`**  
+   - Primary target to run the integration tests
+   - Spins up a clean, dedicated Docker-in-Docker container for the test.  
+   - Builds the project artifacts, creates isolated Docker images, starts the application in that nested container, and then runs end-to-end browser tests.  
+   - Cleans up after itself, ensuring a repeatable, isolated test environment.
```
meiersi-da, line 22: How does this approach support a fast iteration loop, where I fix a few lines of code in the playwright script and rerun it? I haven't tried your approach, but this sounds like we're talking 30s or even a minute of turnaround time for a simple change.

I'd expect a < 10s turnaround time for a good dev experience. This turnaround time really matters!

What we do in the app template is that we use a long-running instance of the whole app, and write the test-code such that it can distinguish steps from different runs: https://github.com/DACH-NY/daml-app-template/blob/8afd5989a206679ae1059548e41766681233c755/app/frontend/e2e-tests/smoke-test.spec.ts#L27

quickstart/Makefile
meiersi-da: > Since there is no API to act on the AppUser's behalf,

@iggy-da : What's missing for the test-runner submit Ledger API command to create the AppInstall request via the JSON API on the user's participant node? Playwright can submit direct HTTP requests: https://playwright.dev/docs/api-testing

docs/development/002-testing.md
hrischuk-da: @meiersi-da you are correct that fast turnaround is needed for the various places in which a code change can happen.  That is an optimization that is not planned for this PR.  This PR was to create a smoke test to make sure that the QS doesn't get broken so fast turnaround is out of scope of this PR.  I'm personally bothered by the startup time of the LocalNet.  I will create a Git issue for this at the different layers to track it.

quickstart/Makefile
iggy-da: Porting the `create-app-install-request` to typescript is possible, I was referring to the existing backend API and application components. Having said that, I think what you are proposing is a good idea, since we are going to port `create-app-install-request` to typescript anyway at some for TypeScript codegen.

I've implemented this in 98609c544e50bcbe8fdeb6a194a8fc51348b3f2f . It is more involved that what one might think due to various authentication and information endpoints that have to be called.

docs/development/002-testing.md
iggy-da: @hrischuk-da is correct, but I think I also understand what you are saying @meiersi-da - you are proposing logical separation of tests from one another using "namespaces" to be able to reuse a continuously running LocalNet instance. I agree that this is a good idea and a good follow-up issue, even though it may not be the most straight-forward to implement in the quickstart context. Furthermore, as Curtis alluded to, these tests are not meant to be run as part of a fast-feedback test-driven-development lifecycle to give live feedback during development, however I take the point that latency is still a consideration.

iggy-da:
@pgarmaz-da has tire-kicked this PR, which included running `make integration-test` 200 times in sequence. In those 200 runs, there was a 0% failure rate, meaning that all tests of all runs passed without flakes.

filmackay (approved):

PR #96, opened 2025-04-09
use ghcr for CN images

dasormeter:
Use ghcr for CN images, as they are now publicly available

Images can be browsed @ https://github.com/orgs/digital-asset/packages?repo_name=decentralized-canton-sync


pgarmaz-da (approved):
Change looks good to me 🎉 That said we'll likely want to have a note in the documentation that informs users that that ghcr.io is the source for some docker images. CC: @hrischuk-da @TheChronicMonster @andrae-da

isegall-da:
ghcr is public for all (as opposed to jfrog where DA was granting access)

PR #107, opened 2025-04-25
docs: update FAQ and upgrades on GS

TheChronicMonster:
Most up to date FAQ
Exploring the Demo now includes "Updates on the GS"


meiersi-da:
FYI @TheChronicMonster : please don't use "GS" as an abbreviation for the Global Synchronizer, as per https://docs.sync.global/faq.html

PR #116, opened 2025-05-11
Create T&Cs for Daml Enterprise Tooling for CN Quickstart.txt

hrischuk-da:


filmackay:
@hrischuk-da suggesting we consider rename this file to something along the lines of `terms.txt`? Having a '&' in the name can cause some heartburn..?

filmackay (approved):

PR #120, opened 2025-05-14
Fix VSCode import of Gradle build and Java projects

adrien-piquerez-da:
The first commit fixes loading the Gradle build in the Gradle for Java extension.

The error was:
```
* Where:
Build file '/Users/adrienpiquerez/github/digital-asset/cn-quickstart2/quickstart/backend/build.gradle.kts' line: 24

* What went wrong:
.env file not found in project root directory (expected at /Users/adrienpiquerez/.vscode/extensions/vscjava.vscode-gradle-3.16.4/lib/.env)
```
That's because the Gradle extension starts Gradle in its own folder, and not in the project root folder. As such it could not find the `.env` file at the project root dir. To fix this we can set a `gradle.root.dir` property and use it instead of `user.dir`.

The second commit fixes the false compilation errors in the Java workspace by declaring the correct source directory for the generated spring Java files.

After this PR I can fully import the Java projects in VSCode, and use code completion and code navigation.

### DIFF
--- a/quickstart/backend/build.gradle.kts
+++ b/quickstart/backend/build.gradle.kts
@@ -115,7 +115,7 @@ sourceSets {
     main {
         java {
             srcDirs(
-                "$projectDir/build/generated-spring/src/main",
+                "$projectDir/build/generated-spring/src/main/java",
                 "$projectDir/build/generated-client/src/main/java",
                 "$projectDir/build/generated-daml-bindings" // TODO: remove this line once daml plugin is used
             )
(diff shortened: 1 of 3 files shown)


peterkvokacka-da:
DId you verify that IntelliJ Idea setup still works?

adrien-piquerez-da:
> DId you verify that IntelliJ Idea setup still works?

I tried this morning, and yes it worked like a charm. Intellisense is working and can even start the backend from within the IDE.

peterkvokacka-da (approved):

PR #121, opened 2025-05-14
Fix Docker env: `.env.local` shoud override `.env`

adrien-piquerez-da:
The values in `.env.local` should override the values in `.env` file.

### DIFF
--- a/quickstart/Makefile
+++ b/quickstart/Makefile
@@ -21,7 +21,7 @@ endif
 
 # Determine the operating system
 KERNEL_NAME := $(shell uname -s)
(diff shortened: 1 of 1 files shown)


peterkvokacka-da:
No problem with that IMHO but why you need this? I haven't been working with the main branch for while so I don't remember what is in .env.local

adrien-piquerez-da:
`.env.local` is a git-ignored file that you can use to customize your local environment without committing the changes. For instance I can change the `DAML_RUNTIME_VERSION` to test a custom runtime.

peterkvokacka-da:
I am working on new version of QS where those files are already in that order you want them just wanted to understand your use case.

PR #130, opened 2025-05-20
Token standard daml model + daml unit test

iggy-da:
- Port daml model to use Token Standard
- Add test suite for daml model happy path

The wallet web UI does not support the Token Standard yet, so end-to-end functionality cannot be tested (manually nor automatically via `make


filmackay-da:
@iggy-da please run through this PR with @peterkvokacka-da as a knowledge handover exercise, so he has a good view of what is in this PR - and an understanding of what is required to be present before this can be merged. ie. without breaking the core workflow, once the Wallet UI is in place.

meiersi-da (commented):
Nice work! This goes in a good direction.

meiersi-da:
> The wallet web UI does not support the Token Standard yet, so end-to-end functionality cannot be tested (manually nor automatically via `make integration-test`).

@iggy-da (FYI: @filmackay-da @peterkvokacka-da ): I'd suggest you to use the Ledger API directly to create the required `Allocation` in the end-to-end test. Thereby retaining test coverage.

Also note that using the Ledger API directly in tests is not an anti-pattern. It's the fallback option for driving workflows that have not (yet) been wrapped up behind higher-level APIs.

quickstart/daml/licensing/daml/Licensing/License.daml
```diff
@@ -1,148 +1,192 @@
...
-    licenseFeeCc : Decimal
+    licenseFeeAmount : Decimal
+    licenseFeeInstrumentId : InstrumentId
     licenseExtensionDuration : RelTime
-    reference : ContractId AppPaymentRequest
-  where 
+    prepareUntil : Time
+    settleBefore : Time
+    createdAt : Time
+    licenseCid : ContractId License
```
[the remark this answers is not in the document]
iggy-da: I've left it like that for now, unless you have a better idea. Reason is that `self` is also not available in `interfaces`, which makes it not obvious what use as a reference cid in the AllocationRequestView -> SettlementInfo Reference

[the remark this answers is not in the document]
meiersi-da: Yes, using the `ContractId License` of the first license is a good idea. To do this in a way that supports updating the license contract, you'll want to copy this construction here: https://github.com/DACH-NY/canton-network-node/blob/ca5448681218a74c3a3fed2837b3da833a610940/token-standard/splice-token-standard-test/daml/Splice/Testing/Apps/TradingApp.daml#L22

Call it something like the `originalLicenseCid : Optional (ContractId License)` and use that one to compare the allocation.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
```diff
@@ -254,37 +257,41 @@ private CompletableFuture<ResponseEntity<Void>> exerciseLicenseRenewChoice(
...
+                                    long payDurationMicros = payDuration.toNanos() / 1_000;
+                                    RelTime paymentAcceptanceDuration = new RelTime(payDurationMicros);
+
+                                    License_Renew choice = new License_Renew(
+                                            new InstrumentId(new Party(dsoResponse.getDsoPartyId()), "Amulet"),
+                                            licenseRenewRequest.getLicenseFeeCc(),
+                                            licenseExtensionDuration,
+                                            paymentAcceptanceDuration,
+                                            Instant.now(),
+                                            // TODO: Make prepareUntil / setUntil offsets configurable
```
peterkvokacka-da, line 278: TODO?

quickstart/daml/licensing/daml/Licensing/AppInstall.daml
```diff
@@ -62,7 +62,7 @@ template AppInstall with
       controller actor
       do
         require "Actor is a signatory" (actor `elem` signatory this)
-        enforceMetadataLimits meta
+        pure ()
```
peterkvokacka-da, line 65: Why this choice does nothing?

quickstart/daml/licensing/daml/Licensing/AppInstall.daml
```diff
@@ -94,10 +92,5 @@ template AppInstallRequest with
...
       controller provider
-      do enforceMetadataLimits meta
-
-    choice AppInstallRequest_Cancel : ()
-      with
-        meta : Metadata -- ^ Machine-readable metadata of the withdrawal.
-      controller user
-      do enforceMetadataLimits meta
+        do
+          pure ()
```
peterkvokacka-da, line 96: Why this choice does nothing?

peterkvokacka-da: ohh it's consuming choice so it just archives the contract?

filmackay-da: > Why this choice does nothing?

It doesn't do nothing - it gives the provider the right to archive the contract, and leaves `meta` evidence of the reason for the rejection and ending of the workflow.

quickstart/daml/licensing/daml/Licensing/AppInstall.daml
filmackay-da: It gives the any individual signatory the right to cancel and end the workflow, again leaving `meta` evidence..

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
iggy-da: Sorry, what is the question?

peterkvokacka-da:
There are no changes in QS UI needed to make this work when we merge it? We need only splice wallet UI to work right?https://github.com/DACH-NY/canton-network-node/milestone/202

quickstart/daml/licensing/daml/Licensing/AppInstall.daml
filmackay-da: > ohh it's consuming choice so it just archives the contract?

Yep - so you were right "do nothing" is kind of it or maybe "leave nothing around"!

iggy-da:
> There are no changes in QS UI needed to make this work when we merge it? We need only splice wallet UI to work right?https://github.com/DACH-NY/canton-network-node/milestone/202

That's essentially right. On a high-level semantic/logical level the workflow is the same. Depending on the final design, there may be backend OpenAPI changes that may need to be propagated to the frontend. The backend SQL queries and code will have to be adapted to the new data-structures (I've made some effort to do so already).

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
peterkvokacka-da: are you going to address the TODO or leave it behind?

iggy-da: Leave behind - can be separate PR

filmackay-da:
@meiersi-da to make sure we're on the same page: this PR will wait until the Wallet UI is available in Splice and then @peterkvokacka-da will integrate it, and make any adjustments that become necessary. 👍

quickstart/daml/licensing/daml/Licensing/License.daml
```diff
@@ -1,148 +1,181 @@
...
-    user : Party
     provider : Party
-    dso: Party
-  deriving (Show, Eq)
+    user : Party
+    dso : Party
+    expiresAt : Time
+    licenseNum : Int
+    params : LicenseParams
+    originalLicenseCid : Optional (ContractId License)
```
meiersi-da, line 35:
```suggestion
    originalLicenseCid : Optional (ContractId License)
      -- ^ The contract-id of the first version of this license contract, which
      -- serves as the unique identifier for the license itself. 
```

meiersi-da (commented):
Thanks a lot for the improvements @iggy-da . There's one major bug for which it would be great to add a test: right now the wrong allocation can be passed in and the license would be extended nevertheless!

Let's fix that. 

Also there's a bunch of minor issues which would be great to fix.

meiersi-da:
> @meiersi-da to make sure we're on the same page: this PR will wait until the Wallet UI is available in Splice and then @peterkvokacka-da will integrate it, and make any adjustments that become necessary. 👍

@filmackay-da : I'm not so keen on adding the dependency on the allocation support in the Wallet UI. In particular not for testing this code end-to-end.

Driving the end-to-end tests of composed applications through the UI's of the dependency seems like a fragile approach wrt upgrades of that dependency (the UI is essentially a private API). 

So I'd suggest to drive the end-to-end test directly via API calls:
1. list holdings via Ledger API
2. call scan to get ChoiceContext for instructing the allocation
3. instruct the allocation via a call to the Ledger API

This approach of creating the allocation follows one-to-one the approach used in the Daml Script tests. 

And with that approach you are unblocked and can work towards merging this PR.

iggy-da:
> not so keen on adding the dependency ... on the Wallet UI. [...] And with that approach you are unblocked and can work towards merging this PR.

Only if `main` is agreed and accepted to be in a red non-working state.

The core workflow happy-path of the application would be broken, even for LocalNet. The user will run into an error when trying to pay for the renewal of a license and won't be able to proceed. I feel it is forgotten who the user of `quickstart` is, and what the name `quickstart` implies, and what the user experience and utility will be if this part of the workflow is broken.

quickstart/daml/licensing/daml.yaml
```diff
@@ -10,6 +10,8 @@ dependencies:
   - daml-stdlib
 data-dependencies:
   # TODO: fetch as artifacts when integrating gradle-daml-plugin https://github.com/DACH-NY/canton-network-node/issues/16393
-- ../dars/splice-amulet-0.1.9.dar
 - ../dars/splice-util-0.1.3.dar
```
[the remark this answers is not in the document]
peterkvokacka-da: hmm we use `require` function from `Splice.Util` do you want us use `assetMsg` instead?

quickstart/daml/licensing/daml/Licensing/License.daml
```diff
@@ -1,148 +1,181 @@
...
+    settleBefore : Time
+    createdAt : Time
+    originalLicenseCid : ContractId License
+  where
     signatory user, provider
 
-    choice LicenseRenewalRequest_Cancel : LicenseRenewalCancel
+    -- | Cancels this renewal request (e.g. if the user rejects or the
+    --   provider withdraws the request).
+    choice LicenseRenewalRequest_Cancel : ()
```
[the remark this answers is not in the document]
peterkvokacka-da: removed

quickstart/daml/licensing/daml.yaml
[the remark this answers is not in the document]
meiersi-da: Just copy the body of this function into your own utils module. This way the app provider keeps control of its evolution. It's a small function.

PR #172, opened 2025-06-19
ensure compatibility of splice version and devnet

### LINKED ISSUE #70: Perform version check on startup
# Background

To avoid confusion/issues/support for users that try to connect to Local/Dev/Net with an incompatible Quickstart, we need to provide some pre-emptive advice or early failure.

# Tasks

1. If running devnet, contact 
https://docs.dev.global.canton.network.sync.global/versions to obtain the version information on Splice
2. Compare this SV Splice version and migration ID against the[ Splice version Quickstart was built with](https://github.com/digital-asset/cn-quickstart/blob/main/quickstart/.env#L9)
3. If remote and local version is exactly the same, proceed; otherwise
4. Warn that the version/migration ID do not match, and display both for the user
5. Confirm the user wants to proceed anyway

erwin-ramirez-da:
closes #70

### DIFF
--- a/quickstart/backend/build.gradle.kts
+++ b/quickstart/backend/build.gradle.kts
@@ -3,6 +3,9 @@
 
 import com.google.protobuf.gradle.*
 import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
+import java.net.URI
+import java.util.Scanner
+import java.nio.file.Files
 
 plugins {
     application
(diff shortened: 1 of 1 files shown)


quickstart/backend/build.gradle.kts
```diff
@@ -72,9 +75,82 @@ tasks.register<Copy>("copyOtelAgentJar") {
 }
 
 tasks.named("build") {
+    dependsOn("validateSpliceCompat")
```
peterkvokacka-da, line 78: we don't want to do this on every build. When I work with localnet I shouldn't care what version devnet is using. I may be behind/after devnet version at any time in feature branch and also in main.

quickstart/backend/build.gradle.kts
```diff
@@ -72,9 +75,82 @@ tasks.register<Copy>("copyOtelAgentJar") {
...
+            ?.version
+            ?: error("Global-Synchronizer-Foundation row not found in devnet")
+
+        logger.lifecycle("""
+            |• DevNet: Splice version: $remoteSpliceVersion
+            |• Local : Splice version: $localVersion
+            """.trimMargin()
+        )
+
+        if (remoteSpliceVersion == localVersion) {
```
peterkvokacka-da, line 109: do this only for devnet. I would also include check whether QS was checked out with correct tag that matches devnet version

quickstart/backend/build.gradle.kts
erwin-ramirez-da: that is what was asked for, plus is not an enforcement, if the user wants to continue with a different version is prompted to continue... the idea as it was presented to me was to inform the user that you are working (dev/local) in a different version than devnet

quickstart/backend/build.gradle.kts
erwin-ramirez-da: see my other comment

peterkvokacka-da: I am proposing change of the requirements here. cc: @filmackay-da The less dependencies we have on devnet the better plus it really doesn't make sense for localnet at all.

quickstart/backend/build.gradle.kts
peterkvokacka-da: That's not correct requirements starts with 

> If running devnet, contact ...

The less dependencies we have on devnet the better plus it really doesn't make any sense for localnet at all and slows developer down on every build for not a good reason.

PR #234, opened 2025-08-27
Rework QS to use TokenStandard

peterkvokacka-da:
**Rework to use** [Token Standard](https://github.com/hyperledger-labs/splice/tree/main/token-standard)  
- Required changes in Daml model, backend-service impl, frontend and e2e tests

**Technical debt cleanup**
- Tracing and logging
    - Introduced a lightweight internal `TracingUtils` API to simplify logging/tracing code.
    - Removed redundant context propagation (`ContextAwareCompletableFutures`); OpenTelemetry instruments CompletableFuture.
    - Trace ID is included in backend-service logs regardless of observability settings.
    - Backend-service logs use the Canton log format to simplify correlation with Canton/Splice logs in lnav: [canton log format](https://github.com/hyperledger-labs/splice/blob/b6086ad6030bbba15db89af823345dd29329008c/canton/canton-json.lnav.json).
- Replaced the non-composable `PQS.activeLeftJoinWhere` with a generic `PQS.query`.
- Added `AuthUtils` with helpers (`asAdminParty`, `asAuthenticatedParty`) to simplify securing the implementations.

**Local development improvements**
- Remote debug for backend-service on port 5005 via DEBUG_ENABLED=true.
- New make target `vite-dev` to run Vite in hot-reload mode while Quickstart is already running.
- Updated Vite welcome screen to link to `app-provider.localhost:5173` (avoids using localhost:5173).
- New make target `restart-service` to restart services cleanly (note: dependent services must be managed separately).
- Fixed PQS setup for `app-user` and `sv`; enable with environment flags (e.g., PQS_APP_USER_PROFILE=on).
- Resolved hanging Docker container issue on failure in create-app-install-request during e2e tests.
- Tracing and logging changes also significantly improve the local development experience.

### DIFF
--- a/quickstart/Makefile
+++ b/quickstart/Makefile
@@ -202,7 +202,7 @@ create-app-install-request: ## Submit an App Install Request from the App User p
 restart-backend: build-backend ## Build and start the application
 	$(call docker-compose, rm -s -f backend-service)
 	$(call docker-compose, rm -s -f register-app-user-tenant)
-	$(call docker-compose, ${DOCKER_COMPOSE_OBSERVABILITY_FILES} $(RESOURCE_CONSTRAINT_CONFIG) up -d --no-recreate)
+	$(call docker-compose, ${DOCKER_COMPOSE_OBSERVABILITY_FILES} $(RESOURCE_CONSTRAINT_CONFIG) up -d --no-recreate register-app-user-tenant)
 
 .PHONY: restart-frontend
 restart-frontend: build-frontend ## Build and start the application
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
@@ -99,7 +99,7 @@ public class LicenseApiImpl implements LicensesApi {
         );
         return auth.asAdminParty(party -> traceServiceCallAsync(ctx, () -> {
             var registryAdminIdFut = tokenStandardProxy.getRegistryAdminId();
-            var licenseFut = damlRepository.findLicenseById(contractId); // KV fix rework to return Optional
+            var licenseFut = damlRepository.findLicenseById(contractId);
             return registryAdminIdFut.thenCombine(licenseFut, (adminId, license) -> {
                 var now = Instant.now();
                 License_Renew choice = new License_Renew(
@@ -171,7 +171,7 @@ public class LicenseApiImpl implements LicensesApi {
                 "commandId", commandId
         );
         return auth.asAuthenticatedParty(party -> traceServiceCallAsync(ctx, () ->
-                damlRepository.findLicenseById(contractId).thenCompose(contract -> { // KV make it return Optional
+                damlRepository.findLicenseById(contractId).thenCompose(contract -> {
                     var license = ensurePresent(Optional.ofNullable(contract), "License not found for contract {}", contractId);
                     var meta = licenseExpireRequest.getMeta().getData();
                     if (!party.equals(auth.getAppProviderPartyId())) {
--- a/quickstart/frontend/src/components/LicenseRenewalRequestModal.tsx
+++ b/quickstart/frontend/src/components/LicenseRenewalRequestModal.tsx
@@ -126,7 +126,8 @@ export default function LicenseRenewalRequestModal({
                   </td>
                   <td className={`ellipsis-cell ${renewal.settleDeadlinePassed && 'deadline-passed'}`}>
                     {formatDateTime(renewal.settleBefore)}
-                    {/* KV date component and remove server side computation of settleDeadlinePassed */}
+                    {/* TODO https://github.com/digital-asset/cn-quickstart/issues/239
+                         date component and remove server side computation of settleDeadlinePassed */}
                   </td>
                   <td className="ellipsis-cell">{renewal.description}</td>
                   <td className="ellipsis-cell">{makeStatus(renewal)}</td>
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/repository/DamlRepository.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/repository/DamlRepository.java
@@ -36,12 +36,12 @@ public class DamlRepository {
         this.pqs = pqs;
     }
 
-    public static record LicenseRenewalRequestWithAllocationCid(
+    public record LicenseRenewalRequestWithAllocationCid(
             Contract<LicenseRenewalRequest> renewal,
             Optional<ContractId<Allocation>> allocationCid) {
     }
 
-    public static record LicenseWithRenewalRequests(
+    public record LicenseWithRenewalRequests(
             Contract<License> license,
             List<LicenseRenewalRequestWithAllocationCid> renewals) {
     }
@@ -114,7 +114,9 @@ public class DamlRepository {
         ).thenApply(v -> new java.util.ArrayList<>(map.values()));
     }
 
-    // KV we should be looking for active ones  only
+    // TODO https://github.com/digital-asset/cn-quickstart/issues/238
+    //  all find functions findById here should look for active contracts only
+    //  and return Optional<Contract<T>>
     /**
      * Fetches a License contract by contract ID.
      */
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/SpringSecurityOAuth2Config.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/SpringSecurityOAuth2Config.java
@@ -95,9 +95,8 @@ public class SpringSecurityOAuth2Config {
                 Collection<GrantedAuthority> authorities = new HashSet<>(defaultGrantedAuthoritiesConverter.convert(jwt));
                 // there is only one AppProvider issuer that can issue JWT to authenticate to ResourceServer
                 // we consider anybody with JWT from that issuer to be admin
-                // KV explain why linking to app config
-                //  add VIRTUAL_PARTY_ID_CLAIM and tenant claims maybe via custom JwtAuthenticationToken  that on getAttributes return also our custom claims
-                //  or maybe check for virtual_partyId claim in authorities instead of claims???
+                // TODO KV https://github.com/digital-asset/cn-quickstart/issues/237
+                //  consider storing partyId/tenantId as GrantedAuthority instead of using JWT claims
                 authorities.add(new SimpleGrantedAuthority("ROLE_ADMIN"));
                 return authorities;
             }
@@ -116,7 +115,7 @@ public class SpringSecurityOAuth2Config {
                 OAuth2AuthorizedClientProviderBuilder.builder()
                         .clientCredentials()
                         .authorizationCode()
-                        .refreshToken() // TODO verify
+                        .refreshToken()
                         .build();
 
         AuthorizedClientServiceOAuth2AuthorizedClientManager authorizedClientManager =
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AdminApiImpl.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AdminApiImpl.java
@@ -68,8 +68,9 @@ public class AdminApiImpl implements AdminApi {
                 "partyId", request.getPartyId()
         );
 
-        // TODO KV fix
-        //  We cannot use auth.asAdminParty(party -> here. if this endpoint is accessed from CLI providing JWT token directly
+        // TODO KV https://github.com/digital-asset/cn-quickstart/issues/237
+        //  consider storing partyId/tenantId as GrantedAuthority instead of using JWT claims
+        //  Problem: We cannot use auth.asAdminParty(party -> here. if this endpoint is accessed from CLI providing JWT token directly
         //  The endpoint is still protected as SpringSecurityOAuth2Config -> HttpSecurity requires role ADMIN for it
         //  but it is annoying that we cannot use the same pattern as in other endpoints
         return traceServiceCallAsync(ctx, () -> CompletableFuture.supplyAsync(() -> {


quickstart/docker/modules/pqs/onboarding/sv/on/shared-secret.env
```diff
@@ -0,0 +1,2 @@
+AUTH_SV_PQS_USER_NAME=ledger-api-user
```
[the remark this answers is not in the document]
peterkvokacka-da: yes that's my next item on cleanup TODO

quickstart/docker/modules/pqs/onboarding/sv/env.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Copyright (c) 2025, Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: 0BSD
+
+set -eo pipefail
+
+source /app/utils.sh
+
+#if [ "$AUTH_MODE" == "oauth2" ]; then
```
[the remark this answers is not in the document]
peterkvokacka-da: yes that's my next item on cleanup TODO

quickstart/frontend/src/components/DurationInput.tsx
```diff
@@ -0,0 +1,99 @@
...
+  inputStyle?: CSSProperties;
+  selectStyle?: CSSProperties;
+  units?: Map<string, string> | string;
+}) {
+  const m = /^(\d+)(.)$/.exec(value || '');
+  const num = m ? Number(m[1]) : 0;
+
+  const defaultLabels: Record<string, string> = {
+    s: 'seconds',
+    m: 'minutes',
```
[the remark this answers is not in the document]
peterkvokacka-da: indeed, thx

quickstart/frontend/src/components/DurationInput.tsx
```diff
@@ -0,0 +1,99 @@
...
+            {label}
+          </option>
+        ))}
+      </select>
+    </>
+  );
+}
+
+export function toISO_8601(duration: string): string {
+  const match = /^(\d+)([smhdM])$/.exec(duration);
```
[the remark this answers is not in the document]
peterkvokacka-da: yeah months should be removed as java.time.Duration supports only subset of  ISO_8601 thus it is pointless for QS to deal with months. But in general if impl choose to support months that is ok as well.

quickstart/frontend/src/components/LicenseRenewalRequestModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+        <button
+          className="btn btn-success btn-issue-renewal"
+          onClick={() => setShowNewModal(true)}
+        >
+          New
+        </button>
+      )}
+
+      <div className="renewals">
+        <table className="table table-fixed xtable-bordered" id="renewals-table">
```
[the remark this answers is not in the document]
peterkvokacka-da: I am ok with empty headers it was like this until now

quickstart/common/openapi.yaml
```diff
@@ -319,16 +319,26 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/CompleteLicenseRenewalRequest'
       responses:
         '200':
           description: License renewed successfully
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/LicenseRenewalResult'
         '400':
```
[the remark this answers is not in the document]
peterkvokacka-da: Happy to address as tech dept issue in separate PR as that is existing problem already in main.

quickstart/docker/modules/pqs/onboarding/sv/env.sh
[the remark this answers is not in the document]
peterkvokacka-da: done

quickstart/docker/modules/pqs/onboarding/sv/on/shared-secret.env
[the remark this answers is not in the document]
peterkvokacka-da: done

quickstart/frontend/src/components/LicenseExpireModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+      title={
+        <div>Archive License</div>
+      }
+      onClose={() => {
+        setExpireDescription('');
+        onClose();
+      }}
+      onConfirm={async () => {
+        if (!expireDescription.trim()) return;
+        await onExpire(expireDescription);
```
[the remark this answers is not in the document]
peterkvokacka-da: Very basic error handling is done in licenseStrore.tsx -> expireLicense

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+    public String getAppProviderPartyId() {
+        return appProviderPartyId;
+    }
+
+    public <T> CompletableFuture<ResponseEntity<T>> asAdminParty(Function<String, CompletableFuture<ResponseEntity<T>>> cf) {
+        return asAuthenticatedParty(party -> {
+            if (party.equals(appProviderPartyId)) {
+                return cf.apply(party);
+            } else {
+                logger.error("Access denied: authenticated party '{}' does not match AppProvider party '{}'.", party, appProviderPartyId);
```
erwin-ramirez-da, line 35: it shouldn't be marked as error, it should be info or max warn

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+                return CompletableFuture.completedFuture(new ResponseEntity<>(HttpStatus.FORBIDDEN));
```
erwin-ramirez-da, line 36: use semantic builders... eg ResponseEntity.status(HttpStatus.FORBIDDEN).build;

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+        });
+    }
+
+    public <T> CompletableFuture<ResponseEntity<T>> asAuthenticatedParty(Function<String, CompletableFuture<ResponseEntity<T>>> cf) {
+        var authParty = authenticatedPartyProvider.getParty();
+        return authParty.map(s -> CompletableFuture.completedFuture(s).thenCompose(cf))
+                .orElseGet(() -> {
+                    logger.error("Authentication failed: no authenticated party present in the security context");
```
erwin-ramirez-da, line 45: same as before it shouldn't be an error log...

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+                    return CompletableFuture.completedFuture(new ResponseEntity<>(HttpStatus.UNAUTHORIZED));
```
erwin-ramirez-da, line 46: same use builder

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class AuthUtils {
+    private final AuthenticatedPartyProvider authenticatedPartyProvider;
+    private final Auth auth;
+    private static final Logger logger = LoggerFactory.getLogger(AuthUtils.class);
+
+    @Value("${application.tenants.AppProvider.partyId}")
+    private String appProviderPartyId;
+
+    AuthUtils(AuthenticatedPartyProvider authenticatedPartyProvider, Auth auth) {
```
erwin-ramirez-da, line 20: it is defaulted to public but for sake of standards constructors should be marked as public

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
```diff
@@ -0,0 +1,57 @@
...
+            } else {
+                logger.error("Access denied: authenticated party '{}' does not match AppProvider party '{}'.", party, appProviderPartyId);
+                return CompletableFuture.completedFuture(new ResponseEntity<>(HttpStatus.FORBIDDEN));
+            }
+        });
+    }
+
+    public <T> CompletableFuture<ResponseEntity<T>> asAuthenticatedParty(Function<String, CompletableFuture<ResponseEntity<T>>> cf) {
+        var authParty = authenticatedPartyProvider.getParty();
+        return authParty.map(s -> CompletableFuture.completedFuture(s).thenCompose(cf))
```
erwin-ramirez-da, line 43: you can apply directly the handler, something like:
```
      return authenticatedPartyProvider.getParty()
                   .map(s -> cf.apply(s))
                   .orElseGet(() -> {
                   ...
```

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseRenewalRequestsApiImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                "commandId", commandId
+        );
+        return auth.asAdminParty(party -> traceServiceCallAsync(ctx, () ->
+                damlRepository.findActiveAllocationRequestById(contractId).thenCompose(allocReq -> {
+                    var allocationRequest = ensurePresent(allocReq, "AllocationRequest {} not found", contractId);
+                    var choice = new AllocationRequest.AllocationRequest_Withdraw(
+                            new ExtraArgs(new ChoiceContext(Map.of()), new Metadata(Map.of()))
+                    );
+                    return ledger.exerciseAndGetResult(allocationRequest.contractId, choice, commandId)
+                            .thenApply(result -> ResponseEntity.ok().build());
```
erwin-ramirez-da, line 61: this is more accurate as a 204 (success with no body) than a 200. can be a tech debt

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseRenewalRequestsApiImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                "contractId", contractId,
+                "commandId", commandId
+        );
+        return auth.asAdminParty(party -> traceServiceCallAsync(ctx, () ->
+                damlRepository.findActiveAllocationRequestById(contractId).thenCompose(allocReq -> {
+                    var allocationRequest = ensurePresent(allocReq, "AllocationRequest {} not found", contractId);
+                    var choice = new AllocationRequest.AllocationRequest_Withdraw(
+                            new ExtraArgs(new ChoiceContext(Map.of()), new Metadata(Map.of()))
+                    );
+                    return ledger.exerciseAndGetResult(allocationRequest.contractId, choice, commandId)
```
erwin-ramirez-da, line 60: ledger.exerciseAndGetResult can throw exception... are we not interested in return proper error message, example what you have done in the ensurePresent method? or happy as it is now

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceException.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.server.ResponseStatusException;
+
+/**
+ * Custom exception class for service layer errors.
+ * Extends ResponseStatusException to include HTTP status codes.
+ * Uses SLF4J's MessageFormatter for message formatting.
+ */
+ class ServiceException extends ResponseStatusException {
+
+    ServiceException(HttpStatus status, String message, Object... args) {
```
erwin-ramirez-da, line 16: marked as public

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
```diff
@@ -62,488 +57,214 @@ public class LicenseApiImpl implements LicensesApi {
...
+                        new InstrumentId(new Party(adminId), "Amulet"),
+                        request.getLicenseFeeCc(),
+                        parseRelTime(request.getLicenseExtensionDuration()),
+                        now,
+                        now.plus(Duration.parse(request.getPrepareUntilDuration())),
+                        now.plus(Duration.parse(request.getSettleBeforeDuration())),
+                        request.getDescription()
                 );
+                return ledger.exerciseAndGetResult(license.contractId, choice, commandId)
+                        .thenApply(r -> ResponseEntity.ok().<Void>build());
```
erwin-ramirez-da, line 118: should be a 204... this is def a tech debt in many places I think

erwin-ramirez-da (commented):
great job tackling so much changes and logging improvement at the same time than implementing ts... few comments added, which I tend to believe will be better address as tech debt... missing in new files the headers from DA

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
peterkvokacka-da: it is not defaulted to public
`'AuthUtils()' is not public in 'com.digitalasset.quickstart.security. AuthUtils'. Cannot be accessed from outside package`

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceException.java
peterkvokacka-da: it is intentionally not public so it cannot be instantiated from outside of service package

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
peterkvokacka-da: existing tech debt in main

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
```diff
@@ -232,9 +223,75 @@ public static void logError(Logger logger, String message, Throwable t) {
...
+        ).thenCompose(f -> f);
+    }
+
+    private static <T> CompletableFuture<T> _trace(
+            TracingUtils.TracingContext ctx,
+            boolean startEvent,
+            Supplier<CompletableFuture<T>> body) {
+        var span = Span.current();
+        if (startEvent) addEventWithAttributes(span, ctx.message() + " start", ctx.attrs());
+        setSpanAttributes(span, ctx.attrs());
```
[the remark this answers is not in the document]
peterkvokacka-da: Span.current() in whenComplete theoretically could be different span then one stored at when method starts. If we use same span is we do now it is obvious that we recordException to the same span that was created when method started.

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
```diff
@@ -197,10 +196,13 @@ public static void logError(Logger logger, String message, Map<String, Object> a
             attributes.forEach(logBuilder::addKeyValue);
         }
         if (t != null) {
+            StringWriter sw = new StringWriter();
+            PrintWriter pw = new PrintWriter(sw);
             logBuilder.setCause(t);
+            t.printStackTrace(pw);
+            logger.error(message + t.getMessage() + "\n" + sw);
```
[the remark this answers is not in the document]
peterkvokacka-da: So what is your suggestion?
Current impl is because previous implementation for some reason didn't print stacktrace at all. If you have better way to do it let me know. It was not added as part of this PR thus I didn't test it or pay attention to it. Diff just shows that it was moved within the file.

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
```diff
@@ -232,9 +223,75 @@ public static void logError(Logger logger, String message, Throwable t) {
...
+        logError(logger, message, null, null);
+    }
+
+    public static TracingContext tracingCtx(Logger logger, Object... args) {
+        if (args == null || args.length % 2 == 0) {
+            throw new IllegalArgumentException("attrs requires an odd number of arguments name, (key, value pairs).");
+        }
+        Map<String, Object> map = new HashMap<>();
+        for (int i = 1; i < args.length; i += 2) {
+            String key = args[i] == null ? "null" : args[i].toString();
```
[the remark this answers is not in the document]
peterkvokacka-da: Yes some cleanup is needed here. Function accepts Object as implicit conversion to string didn't happen all the time which was annoying. So `traceCtx` ensure conversion by calling toString().
Expected arguments are message + keyvalue pairs. Message was not introduced as separate argument purely for formatting reason in IJ. Which I will admit is not strong reason so if you insist I can change that.

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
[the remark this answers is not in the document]
erwin-ramirez-da: yeah use the .`setMessage` method... that should do

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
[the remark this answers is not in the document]
erwin-ramirez-da: problem is that if you used in the .whenComplete it could mean it is closed... it is not happening now as you have the autowire but that is the right way to do so

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
peterkvokacka-da: So what qualifies as error than? Because you can say exactly the same about any error. "That is not an error... it is license not found ...."
Not that it would matter in this case as in reality you will not see this `Access denied error` ever, unless you messed up spring boot http security config, as spring boot should not allow you to call the endpoint in the first place. The check here is just insurance policy against spring boot security misconfiguration.

erwin-ramirez-da: Well I think the answer is given by the return type you are giving back... you are not returning an exception (like normally it would be ResponseStatusException), you are returning something like: 

`CompletableFuture.completedFuture(new ResponseEntity<>(HttpStatus.FORBIDDEN))`

If you consider that the non-authentication is and error (which is valid, just need to be standardised everywhere), the proper thing to do is to do something like:

 `throw new ResponseStatusException(HttpStatus.FORBIDDEN, "Access denied: authenticated party does not have admin privileges.");`

This doesn't change the signature of the method neither what the user will receive, but is more consistent to what you want.

But again for me authorisation detecting that is not authorised is a normal part of the auth flow... but if you consider that is not, just change the return type on the unauthorised logic.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceException.java
erwin-ramirez-da: got ya... then maybe worth to make the class final??

quickstart/Makefile
```diff
@@ -198,6 +208,14 @@ restart-backend: build-backend ## Build and start the application
 restart-frontend: build-frontend ## Build and start the application
 	$(call docker-compose, restart nginx)
 
+.PHONY: restart-service
+restart-service:
+	@if [ -z "${SERVICE}" ]; then \
+	  echo "Usage: make restart-service SERVICE=<service>"; exit 1; \
+	fi
+	$(call docker-compose, rm -s -f ${SERVICE})
+	$(call docker-compose, up -d --no-recreate)
```
amarqueslee-da, line 217: Should this maybe be:

```suggestion
	$(call docker-compose, up -d --no-recreate ${SERVICE})
```

That is, bring up the service you `rm` d rather than all the services?

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
```diff
@@ -562,10 +283,25 @@ private static ValueOuterClass.Identifier parseTemplateIdentifier(String templat
...
+
+    private static RelTime parseRelTime(String durationStr) {
+        Duration duration = Duration.parse(durationStr);
+        long micros = duration.toNanos() / 1_000;
+        return new RelTime(micros);
+    }
+
+    private static OffsetDateTime toOffsetDateTime(Instant instant) {
+        return instant == null ? null : OffsetDateTime.ofInstant(instant, ZoneOffset.UTC);
+    }
```
amarqueslee-da, line 302: These don't look awfully specific to the licence API implementation. `toAnyValueContractId` also looks like it can be `static`.

quickstart/backend/src/main/java/com/digitalasset/quickstart/security/AuthUtils.java
peterkvokacka-da: How the response for HTTPRequest is constructed that's just technical formality (return type or exception) I was keen to learn your reasoning about what error you consider should be logged on WARN and what on ERROR level.

Btw return type is as it is to support function chaining in the controller's function implementation changing the return type would make it inconvenient to use. Furthermore we need to make sure the function is called on HTTP Request Handler thread and signature that allows function chaining is encouraging you to do so.

quickstart/backend/src/main/java/com/digitalasset/quickstart/utility/TracingUtils.java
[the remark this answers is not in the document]
peterkvokacka-da: Just tested. No need to capture stacktrace anymore. Will removing it.

PR #240, opened 2025-09-02
added identification to docker process to include them in the clean-all task 

### LINKED ISSUE #170: Rework make clean-all
Change `make clean-all` so it works correctly even when user changes setup via `make setup` whilst previously started containers are still running and then runs `make clean-all`

erwin-ramirez-da:
closes #170 
added labels to all the possible containers volumes and networks generated in the startup of the quickstart so it can be identified and removed in case of clean-all command


peterkvokacka-da:
We should not modify `quickstart/docker/modules/localnet` as it is copy of Splice LocalNet from the splice repo.
What about run what is currently in make clean-docker (docker compose down -v) and then if there are still services running under `quickstart` network ask  user for permission to remove them and do
```
docker ps -q --filter network=quickstart | xargs -r docker rm -f && docker network rm quickstart
```
?
This way we don't need to care about adding label to any new docker service we add in a future.

PR #252, opened 2025-09-12
Issue 236 improve standards in open api specs

### LINKED ISSUE #236: Modify openAPI specifications to properly use http codes and manage error handling appropriately
Most of the end points described in the OpenAPI are not appropriately mapped to what the standard http status codes are designed (eg. created should be 201), as well and more important the errors send by the backend are lacking of interpretation (eg. most of the current entries are described as returning 200, but few specify the possibly errors coming from the backend as 400 Bad request, or 409 Conflict).... this is important mostly to present user a proper info when things goes wrong in terms of the different flows in the Quickstart

### LINKED ISSUE #238: Clean DamlRepository findAaaById
all findById functions in DamlRepository should look for active contracts only
and return Optional<Contract<T>>

erwin-ramirez-da:
closes #236 
closes #238 

Updating definitions and implementations of the endpoints used in the user flow in the QS

### DIFF
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/pqs/Pqs.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/pqs/Pqs.java
@@ -82,16 +82,15 @@ public class Pqs {
     }
 
     /**
-     * Retrieves a single active contract of a specific template type matching a
-     * custom WHERE clause.
+     * Retrieves a contract by its contract ID from the underlying store.
      */
     @WithSpan
-    public <T extends Template> CompletableFuture<Optional<Contract<T>>> activeContractByContractId(
+    public <T extends Template> CompletableFuture<Optional<Contract<T>>> contractByContractId(
             Class<T> clazz,
             Object... params
     ) {
         Identifier identifier = Utils.getTemplateIdByClass(clazz);
-        var ctx = tracingCtx(logger, "PQS activeContractByContractId",
+        var ctx = tracingCtx(logger, "PQS contractByContractId",
                 "templateId", identifier.qualifiedName(),
                 "params", params
         );
@@ -100,29 +99,6 @@ public class Pqs {
             try {
                 return Optional.ofNullable(jdbcTemplate.queryForObject(sql, new PqsContractRowMapper<>(identifier),
                     combineParams(identifier.qualifiedName(), params)));
-            } catch (EmptyResultDataAccessException e) {
-                return Optional.empty();
-            }
-        });
-    }
-
-    /**
-     * Retrieves a contract by its contract ID from the underlying store.
-     */
-    @WithSpan
-    public <T extends Template> CompletableFuture<Optional<Contract<T>>> byContractId(
-            Class<T> clazz,
-            String id
-    ) {
-        Identifier identifier = Utils.getTemplateIdByClass(clazz);
-        var ctx = tracingCtx(logger, "byContractId",
-                "templateId", identifier.qualifiedName(),
-                "contractId", id
-        );
-        return runAndTraceAsync(ctx, () -> {
-            String sql = "select contract_id, payload from lookup_contract(?, ?)";
-            try {
-                return Optional.ofNullable(jdbcTemplate.queryForObject(sql, new PqsContractRowMapper<>(identifier), id, identifier.qualifiedName()));
             } catch (EmptyResultDataAccessException e) {
                 return Optional.empty();
             }
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
@@ -138,9 +138,7 @@ public class LicenseApiImpl implements LicensesApi {
             var choiceContextFut = tokenStandardProxy.getAllocationTransferContext(request.getAllocationContractId());
             var renewalFut = damlRepository.findActiveLicenseRenewalRequestById(request.getRenewalRequestContractId());
             return choiceContextFut.thenCombine(renewalFut, (c, r) -> {
-                var choiceContext = c.orElseThrow(
-                    () -> new ResponseStatusException(HttpStatus.NOT_FOUND, String.format("Transfer context not found for allocation %s", request.getAllocationContractId()))
-                );
+                var choiceContext = ensurePresent(c, "Transfer context not found for allocation %s", request.getAllocationContractId());
                 var renewal = ensurePresent(r, "Active renewal request not found for contract %s", request.getRenewalRequestContractId());
                 TransferContext transferContext = prepareTransferContext(
                         choiceContext.getDisclosedContracts(),
--- a/quickstart/frontend/src/stores/tenantRegistrationStore.tsx
+++ b/quickstart/frontend/src/stores/tenantRegistrationStore.tsx
@@ -9,7 +9,7 @@ import type {
     TenantRegistration,
     TenantRegistrationRequest
 } from '../openapi.d.ts';
-import {withErrorHandling} from "../utils/error.tsx";
+import {withErrorHandling} from "../utils/error";
 
 
 interface TenantRegistrationState {
--- a/quickstart/backend/src/main/java/com/digitalasset/quickstart/repository/DamlRepository.java
+++ b/quickstart/backend/src/main/java/com/digitalasset/quickstart/repository/DamlRepository.java
@@ -114,26 +114,33 @@ public class DamlRepository {
         ).thenApply(v -> new java.util.ArrayList<>(map.values()));
     }
 
-    // TODO https://github.com/digital-asset/cn-quickstart/issues/238
-    //  all find functions findById here should look for active contracts only
-    //  and return Optional<Contract<T>>
     /**
      * Fetches a License contract by contract ID.
      */
     public CompletableFuture<Optional<Contract<License>>> findLicenseById(String contractId) {
-        return pqs.byContractId(License.class, contractId);
-    }
-
-    public <T extends Template> CompletableFuture<Optional<Contract<T>>> findActive(Class<T> clazz, String contractId) {
-        return pqs.activeContractByContractId(clazz, contractId);
+        return pqs.contractByContractId(License.class, contractId);
     }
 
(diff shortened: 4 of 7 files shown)


quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java
```diff
@@ -113,8 +115,9 @@ public CompletableFuture<ResponseEntity<Void>> renewLicense(
                         request.getDescription()
                 );
                 return ledger.exerciseAndGetResult(license.contractId, choice, commandId)
-                        .thenApply(r -> ResponseEntity.ok().<Void>build());
-            }).thenCompose(x -> x);
+                        .<ResponseEntity<Void>>thenApply(r -> ResponseEntity.created(URI.create(String.format("/licenses/%s:renew", contractId))).build());
```
peterkvokacka-da, line 118: the URI provided is strange. Ideally it should be "/licenses/" + contractId but we do not support that. I would suggest removing URI completely as it doesn't make sense as it is right now. Same applies to other endpoints.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceUtils.java
```diff
@@ -28,6 +28,16 @@ static <T> CompletableFuture<T> traceServiceCallAsync(
             Supplier<CompletableFuture<T>> body) {
         return TracingUtils.traceWithStartEventAsync(ctx, body);
     }
+
+    static Throwable mapToHttp(Throwable throwable) {
+        //  maintain the original exception if it is already a ResponseStatusException
+        if (throwable instanceof ResponseStatusException rse) return rse;
+        //  this error is from the findById lookup
+        if (throwable instanceof org.springframework.dao.EmptyResultDataAccessException)
```
peterkvokacka-da, line 36: This is tight coupling to spring DAO. We should do this https://github.com/digital-asset/cn-quickstart/issues/238 instead and take care of the error like it is already done in `completeLicenseRenewal` then `mapToHttp` won't be needed.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AppInstallsApiImpl.java
```diff
@@ -121,19 +127,20 @@ public CompletableFuture<ResponseEntity<Void>> cancelAppInstall(
...
+                                && !party.equals(contract.payload.getProvider.getParty)) {
+                            throw new ResponseStatusException(HttpStatus.FORBIDDEN, String.format("party %s is not the user nor provider", party));
+                        }
+                        Metadata meta = new Metadata(appInstallCancel.getMeta().getData());
+                        // topologically we can only act as the provider
+                        Party provider = new Party(auth.getAppProviderPartyId());
+                        AppInstall_Cancel choice = new AppInstall_Cancel(provider, meta);
+                        return ledger.exerciseAndGetResult(contract.contractId, choice, commandId)
+                                .thenApply(result -> ResponseEntity.noContent().build());
+                }).exceptionallyCompose(ex -> CompletableFuture.failedFuture(mapToHttp(ex)))
```
peterkvokacka-da, line 143: whatever `mapToHttp` is doing it could be done in one place in `ServiceUtils.traceServiceCallAsync`

quickstart/common/openapi.yaml
```diff
@@ -59,68 +94,58 @@ paths:
...
+                  walletUrl: "http://wallet.localhost:2000/"
+                  clientId: "client-xyz"
+                  issuerUrl: "http://keycloak.localhost:8082/realms/AppProvider"
+                  internal: false
+              sharedSecret:
+                value:
+                  tenantId: "tenant-ss"
+                  partyId: "party-ss"
+                  walletUrl: "http://wallet.localhost:2000/"
+                  users: [ "alice@example.com","bob@example.com" ]
```
peterkvokacka-da, line 111: this should be just `alice, bob`

peterkvokacka-da (changes requested):
Lots of errors (with different error statuses) on backend service impl were introduced in this PR, it would make sense then handle them properly in the frontend. 
There is a different proposal https://github.com/digital-asset/cn-quickstart/issues/238 for NOT_FOUND error handling, while keeping our components loosely coupled.

erwin-ramirez-da:
> Lots of errors (with different error statuses) on backend service impl were introduced in this PR, it would make sense then handle them properly in the frontend. There is a different proposal #238 for NOT_FOUND error handling, while keeping our components loosely coupled.

I thought about same idea, but after reading realise this proposal is base on Functional programming (which I agree on), but the Java style and its standards (we use the JdbcTemplate from springframework) is to follow try/catch system. I think we should stick to the idea that we are in Java world and get ourselves into the idea of manage Java style as users will be more use to Java style than FT style.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceUtils.java
erwin-ramirez-da: This is a way to go into FT style programming in the world of Java (I am not against of it), but we need to think this is going to be used for Java programers.
I am happy to go in this direction but then we will not follow standards for example for the JdbcTemplate of spring framework of try/catch.
I am just concerned that users are Java use to style rather than FT style.

peterkvokacka-da: FT style or not the main problem is tight coupling.

peterkvokacka-da:
> > Lots of errors (with different error statuses) on backend service impl were introduced in this PR, it would make sense then handle them properly in the frontend. There is a different proposal #238 for NOT_FOUND error handling, while keeping our components loosely coupled.
> 
> I thought about same idea, but after reading realise this proposal is base on Functional programming (which I agree on), but the Java style and its standards (we use the JdbcTemplate from springframework) is to follow try/catch system. I think we should stick to the idea that we are in Java world and get ourselves into the idea of manage Java style as users will be more use to Java style than FT style.

You can catch them all in the proper layer DamlRepository or PQS and deal with them in your preferred way either FT style or not. Fronted should be adjusted anyway.

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceUtils.java
```diff
@@ -26,7 +26,19 @@ static <T> T ensurePresent(Optional<T> opt, String message, Object... args) {
     static <T> CompletableFuture<T> traceServiceCallAsync(
             TracingUtils.TracingContext ctx,
             Supplier<CompletableFuture<T>> body) {
-        return TracingUtils.traceWithStartEventAsync(ctx, body);
+        return TracingUtils.traceWithStartEventAsync(ctx, body)
+                   .exceptionallyCompose(ex -> CompletableFuture.failedFuture(mapToHttp(ex)))
+            ;
```
peterkvokacka-da, line 31:
```suggestion
        return TracingUtils.traceWithStartEventAsync(
            ctx.exceptionallyCompose(ex -> CompletableFuture.failedFuture(mapToHttp(ex))),
            body
         );
```

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AppInstallsApiImpl.java
```diff
@@ -90,8 +91,10 @@ public CompletableFuture<ResponseEntity<AppInstallCreateLicenseResult>> createLi
                 "commandId", commandId
         );
         return auth.asAdminParty(party -> traceServiceCallAsync(ctx, () ->
-                damlRepository.findAppInstallById(contractId).thenCompose(contract -> {
+                damlRepository.findAppInstallById(contractId).thenComposeAsync(contract -> {
                     String providerParty = contract.payload.getProvider.getParty;
+                    // TODO Peter: Check with Peter if this provider party is not the same checked in asAdminParty
```
peterkvokacka-da, line 96: it is the same but in asAdminParty() we are checking whether you have access to this endpoint as admin.
The check on line 98 is checking that provider on the contract is for this providerParty (we have only one in QS and admin is actually provider)

quickstart/backend/src/main/java/com/digitalasset/quickstart/service/ServiceUtils.java
```diff
@@ -26,7 +26,19 @@ static <T> T ensurePresent(Optional<T> opt, String message, Object... args) {
...
+    }
+
+    private static Throwable mapToHttp(Throwable throwable) {
+        //  maintain the original exception if it is already a ResponseStatusException
+        if (throwable instanceof ResponseStatusException rse) return rse;
+        //  this error is from the findById lookup
+        if (throwable instanceof org.springframework.dao.EmptyResultDataAccessException)
```
peterkvokacka-da, line 38: IMHO This very likely will not work as intended as `mapToHttp` is currently used
```
cf.exceptionallyCompose(ex -> CompletableFuture.failedFuture(mapToHttp(ex)))
```
so the `EmptyResultDataAccessException` is nested in `CompletionException`. Did you test it?

erwin-ramirez-da: yes I did, it works as expected... but I am changing regardless...

quickstart/frontend/src/stores/tenantRegistrationStore.tsx
```diff
@@ -4,27 +4,14 @@
 import React, { createContext, useContext, useState, useCallback } from 'react'
 import { useToast } from './toastStore'
 import api from "../api.ts";
+import type {
+    Client,
+    TenantRegistration,
+    TenantRegistrationRequest
+} from '../openapi.d.ts';
+import {withErrorHandling} from "../utils/error.tsx";
```
peterkvokacka-da, line 12: name it `error.ts` please as that is pure typescript file without any react

quickstart/backend/src/main/java/com/digitalasset/quickstart/pqs/Pqs.java
```diff
@@ -87,30 +86,31 @@ public <T extends Template> CompletableFuture<List<Contract<T>>> activeWhere(
      * custom WHERE clause.
```
peterkvokacka-da, line 86: Wrong comment as you removed the `whereClause` argument.
Can we actually remove this function completely in favour of `byContractId`? It seems they are doing the same thing?

PR #261, opened 2025-09-22
prefilling party Id for tenant creation

### LINKED ISSUE #244: Use a default a party hint instead of asking
for make setup, can we stop asking the user to specify a party hint and just print out one that was selected.  This data is not needed if the LocalNet doesn't connect with DevNet.  It simplifies the UX.

erwin-ramirez-da:
pre-filing  partyId in the creation of tenant flow
close #244

### DIFF
--- a/quickstart/frontend/src/views/TenantRegistrationView.tsx
+++ b/quickstart/frontend/src/views/TenantRegistrationView.tsx
@@ -2,14 +2,11 @@
 // SPDX-License-Identifier: 0BSD
 
 import React, { useEffect, useState } from 'react'
-import {
-    useTenantRegistrationStore
-} from '../stores/tenantRegistrationStore'
+import { useTenantRegistrationStore } from '../stores/tenantRegistrationStore'
 import type { TenantRegistrationRequest } from "../openapi.d.ts"
-import {useToast} from '../stores/toastStore';
-import api from '../api';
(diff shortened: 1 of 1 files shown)


quickstart/frontend/src/views/TenantRegistrationView.tsx
```diff
@@ -28,29 +28,36 @@ const TenantRegistrationView: React.FC = () => {
...
-    };
+    }
 
     useEffect(() => {
-        fetchFeatureFlags();
+        fetchFeatureFlags()
         fetchTenantRegistrations()
     }, [fetchTenantRegistrations])
 
+    // Prefill Party ID (unchangeable)
```
peterkvokacka-da, line 46: why we are doing all of this? the #244 didn't ask for it and it seems wrong to me.
IIUC #244 asked only for changes in `make setup`. It didn't ask for changes in tenants tab.
I would remove UI changes and keep `make setup` changes only.
@hrischuk-da can you confirm?

peterkvokacka-da (changes requested):
Looks like a few bonus changes made their way in 😅

quickstart/frontend/src/views/TenantRegistrationView.tsx
hrischuk-da: @peterkvokacka-da are you objecting to anything in particular?

PR #265, opened 2025-09-23
docs: capture-logs date or gdate

miller-da:
Using existing kernel check in `Makefile` to get appropriate date command (`date` for Linux or `gdate` for MacOS). This resolves a warning on Macs when using `make capture-logs`.

Adds two new FAQs

### DIFF
--- a/quickstart/Makefile
+++ b/quickstart/Makefile
@@ -7,13 +7,6 @@
 # Determine the operating system
 KERNEL_NAME := $(shell uname -s)
 
-# Set appropriate date command based on OS
-ifeq ($(KERNEL_NAME),Darwin)
-    DATE_CMD = gdate
-else
-    DATE_CMD = date
-endif
-
 ifneq (,$(wildcard .env))
     include .env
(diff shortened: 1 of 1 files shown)


TheChronicMonster (approved):

peterkvokacka-da:
interesting I am on MacOS and don't see these illegal option. Thank you for addressing that.

quickstart/Makefile
```diff
@@ -7,6 +7,13 @@
 # Determine the operating system
 KERNEL_NAME := $(shell uname -s)
 
+# Set appropriate date command based on OS
+ifeq ($(KERNEL_NAME),Darwin)
+    DATE_CMD = gdate
+else
+    DATE_CMD = date
+endif
+
```
pgarmaz-da, line 16: If we are going to use nix to provide `date` via the `coreutils` package, then we shouldn't need to flip between `gdate` and `date` based on the OS.  Further more I don't have `gdate` available on my system.

PR #272, opened 2025-09-25
docs: canton console connections

miller-da:


peterkvokacka-da (approved):
LGTM
we may later introduce more convenient aliases in Splice LocalNet

PR #291, opened 2025-11-10
docs: bump directory tree in quickstart guide

jaspervdj-da:
This make it (more) clear that the daml/, backend/ and frontend/ directories should be nested under myapp/.


dasormeter (dismissed):

dasormeter:
> @jaspervdj-da , @dasormeter it seems like the ci test is stuck. I don't see a way to access and restart it.

yeah i think something in the diff logic is not seeing this docs only change as triggering the ci run -- needs some investigation to see how to kick it to life

dasormeter:
oh @jaspervdj-da -- i think this is not running CI because you have contributed from a fork

Can you try just checking out the repo and creating a standard branch?

PR #320, opened 2026-02-17
Add Vagrantfile for local development

giner:
This makes it possible to have a VM with all necessities quickly app and running as well as an easy way to have a clean environment at any moment.

### DIFF
--- a/README.vagrant.md
+++ b/README.vagrant.md
@@ -1,0 +1,64 @@
+# Vagrantfile for local development environment
+
+This Vagrantfile sets up a local development environment for by creating a
+virtual machine with Nix, direnv and a few other tools pre-installed.
+
+## Requirements
+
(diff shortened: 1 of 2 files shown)


dasormeter:
@giner why don't you raise this is a community-contributed enhancement under 
https://docs.dev.sync.global/community/index.html?

Rgds

giner:
@dasormeter, thank you, I didn't know about the community contributions page. I've given it a thought and I think that having `Vagrantfile` for `cn-quickstart` on that page is unlikely to serve the purpose as it'll be hard to spot, inconvenient to use and maintain. If it's in the repo all you do is cloning the repo and running `vagrant up`. I'm maintaining a similar one for splice: https://github.com/hyperledger-labs/splice/blob/main/Vagrantfile

PR #347, opened 2026-05-05
feat(backend-js): add a Node.js implementation of the Quickstart backend

### LINKED ISSUE #156: Add a Node.js backend example to the QS
TypeScript and Javascript have very large development communities.  Adding an example backend that runs in Node.js is appropriate.

As a proof point, I just had a developer from Deploi request this.  He said it would have spared him hours of reading and trial/error.

fernandomg:
Closes #156

## Contributed by

[@BootNodeDev](https://github.com/BootNodeDev) — [bootnode.dev](https://bootnode.dev)

## Summary

- Adds a Node.js / TypeScript / Fastify backend that mirrors the Java backend's HTTP API surface, selectable at build time via `BACKEND=js` (persisted in `.env.local` by `make setup`).
- Both backends share the same OpenAPI contract, env files, onboarding volume, listening port, and cookie/CSRF auth surface (OAuth2 and shared-secret). The JS backend talks to Canton via the JSON Ledger API on port `3975` instead of gRPC, reads PQS Postgres directly, and uses the dpm `codegen-js` Daml bindings for typed command construction. The Compose stack swaps in the JS service via an override file when `BACKEND=js`.

## Test plan

- [x] Ran `make clean-all && make setup` and answered `js` to the new interactive **Backend** prompt — alongside the existing prompts for `AUTH_MODE`, `OBSERVABILITY_ENABLED`, `TEST_MODE`, and `PARTY_HINT`. The selection is written to `.env.local` as `BACKEND=js`, so every subsequent `make start` / `make restart-backend` / `make stop` automatically targets the JS service without needing to repeat the flag. Re-ran `make setup` and chose `java` to confirm the prompt round-trips and that the stack swaps back to the Java backend cleanly. Verified under both `AUTH_MODE=oauth2` and `AUTH_MODE=shared-secret`.
- [x] Walked through the entire flow described in the *Explore the Demo* guide (https://docs.digitalasset.com/build/3.5/quickstart/operate/explore-the-demo.html): App User and App Provider login, tenant pre-registration via `register-app-user-tenant`, `AppInstallRequest` accept/reject, `License` issuance, the renewal flow with wallet allocation, and license expiry. Each step produced behaviour identical to the Java backend.

## Deferred to follow-up

The items below were surfaced during review and consciously left out of this PR. Each one is either parity-equal with the Java backend (so a JS-only fix would create drift) or a production-hardening concern that needs a cross-cutting design touching both backends. They should be tracked in a follow-up issue against both implementations.

| # | Issue | Why deferred |
|---|---|---|
| 1 | `SESSION_SECRET` falls back to random bytes per restart in [`backend-js/src/auth/session.ts`](https://github.com/BootNodeDev/cn-quickstart/blob/js-parity/quickstart/backend-js/src/auth/session.ts), invalidating signed session cookies. | Java's in-memory `HttpSession` produces the same observable behaviour: every restart logs all users out (signature mismatch in JS, lost server-side state in Java). Fix needs a stable signing secret **and** a persistent session store applied to both backends. |
| 2 | Raw Canton ledger error text is forwarded to HTTP clients (party IDs, template IDs, choice arguments). | Java's [`application.yml`](https://github.com/BootNodeDev/cn-quickstart/blob/js-parity/quickstart/backend/src/main/resources/application.yml) explicitly enables `include-message: always`, `include-stacktrace: always`, `include-exception: true` for dev — leaks at least as much, with stacktraces. A sanitising error mapper needs to land on both backends together. |
| 3 | Malformed duration strings throw → HTTP 500 where 400 would be the correct status. | Same behaviour in Java: `Duration.parse` throws `DateTimeParseException` and there is no `@ControllerAdvice` mapping it to 400. Fix on both sides in one PR. |
| 4 | `DELETE /admin/tenant-registrations/:tenantId` removes the OAuth2 registry entry **before** the tenant repo entry; if step 2 fails the registry is gone but the tenant remains. | Java's [`deleteTenantRegistration`](https://github.com/BootNodeDev/cn-quickstart/blob/js-parity/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AdminApiImpl.java) uses the same ordering and has the same risk window. Proper transactional/rollback handling needs to land on both backends. |
| 5 | The 201 response body for `AppInstallRequest_Accept` does not include the new AppInstall's `contractId`. | Same omission in Java's [`AppInstallRequestsApiImpl#acceptAppInstallRequest`](https://github.com/BootNodeDev/cn-quickstart/blob/js-parity/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/AppInstallRequestsApiImpl.java). The OpenAPI response shape and the choice-result extraction need to be updated together on both sides. |

## JS-only follow-ups
Unlike the deferred items above (which the JS backend shares with Java and are best fixed on both sides together) the following is a JS-only refinement that wouldn't change behaviour or create drift with Java, so it's tracked separately:

- Type the licensing read path via the generated bindings. The mappers in `mappers.ts` read raw JSON Ledger API / PQS payloads through untyped helpers (`str`/`num`/`meta` over `Record<string, unknown>`). `dpm codegen-js` only emits Daml-shaped types, so the OpenAPI response shaping and derived fields stay hand-written regardless; but the raw payloads could be decoded through the generated `Serializable` bindings for type-safe field access, bringing the read path closer to Java's `LicenseApiImpl`, which maps by hand but reads from generated typed objects.


quickstart/backend-js/src/auth/admin-gate.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { BackendConfig } from '../config.js'
+import { verifyAdminJwt } from './jwt-admin.js'
+
+const hasBearerToken = (authHeader: string | undefined): boolean =>
+  typeof authHeader === 'string' && /^Bearer\s+\S+/i.test(authHeader)
+
+// Accepts either a valid Bearer JWT or an admin session cookie.
+// On failure, replies with 401 (no auth) or 403 (authenticated but not admin) and returns false.
+export const checkAdmin = async (cfg: BackendConfig, req: FastifyRequest, reply: FastifyReply): Promise<boolean> => {
+  const sessionIsAdmin = req.session.user?.isAdmin
```
dasormeter, line 11: how do you determine this `isAdmin` check?

quickstart/backend-js/src/auth/cookies.ts
```diff
@@ -0,0 +1,3 @@
+export const SESSION_COOKIE = 'JSESSIONID'
```
dasormeter, line 1: why are we relying on JSESSIONID?

quickstart/backend-js/src/auth/oauth2.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const registerOAuth2 = async (app: FastifyInstance, cfg: BackendConfig, registry: OAuth2Registry, tenants: TenantRepository): Promise<void> => {
+  if (cfg.authMode !== 'oauth2') return
+
+  app.get<{ Params: { registrationId: string } }>('/oauth2/authorization/:registrationId', async (req, reply) => {
+    const entry = registry.get(req.params.registrationId)
+    if (entry === undefined) { reply.code(404); return { message: 'unknown_registration' } }
+    const codeVerifier = openid.randomPKCECodeVerifier()
+    const codeChallenge = await openid.calculatePKCECodeChallenge(codeVerifier)
+    const state = randomBytes(16).toString('hex')
```
dasormeter, line 31: what are these seedings for?

quickstart/backend-js/src/auth/session.ts
```diff
@@ -0,0 +1,22 @@
...
+  interface Session {
+    user?: { name: string; tenantId: string; partyId: string; userId?: string; idTokenClaims?: Record<string, unknown>; roles?: string[]; isAdmin?: boolean }
+    oauthState?: { state: string; codeVerifier: string; registrationId: string; nonce: string }
+  }
+}
+
+export const registerSession = async (app: FastifyInstance): Promise<void> => {
+  await app.register(cookie)
+  await app.register(session, {
+    secret: process.env['SESSION_SECRET'] ?? randomBytes(32).toString('hex'),
```
dasormeter, line 17: why must SESSION_SECRET be in env var?

quickstart/backend-js/src/domain/licensing/mappers.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const mapAppInstallRequest = (contractId: string, payload: P) => ({
+  contractId,
+  provider: str(payload, 'provider'),
+  user: str(payload, 'user'),
+  meta: meta(payload, 'meta')
+})
+
+export const mapAppInstall = (contractId: string, payload: P) => ({
```
dasormeter, line 19: Can this code be gen'd from the daml to typescript using dpm codegen-js?

quickstart/backend-js/src/auth/admin-gate.ts
fernandomg: The `isAdmin` flag is set at login and the gate just reads it back. In oauth2 mode a user is admin when their registration is the AppProvider ([oauth2.ts:78](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/oauth2.ts#L78)), and in shared-secret mode when the matched tenant is internal ([shared-secret.ts:20](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/shared-secret.ts#L20)).

Beyond that session flag, the gate also lets through a Bearer JWT that's validly signed by the configured issuer ([admin-gate.ts:17](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/admin-gate.ts#L17) / [jwt-admin.ts:22](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/jwt-admin.ts#L22)).

We leaned on the Java backend's model so the two stay in step. It gives ROLE_ADMIN to the AppProvider login ([OAuth2AuthenticationSuccessHandler.java:65](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/oauth2/OAuth2AuthenticationSuccessHandler.java#L65), [SharedSecretConfig.java:108](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/sharedsecret/SharedSecretConfig.java#L108)) and treats any issuer-signed JWT as admin ([OAuth2Config.java:108](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/oauth2/OAuth2Config.java#L108)).

quickstart/backend-js/src/auth/cookies.ts
fernandomg: Nothing special about the name here; JSESSIONID is just what we call the session cookie, kept the same as the Java backend.

We set it here (applied in [session.ts:18](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/session.ts#L18)), and Spring uses the same name and clears it on logout ([OAuth2Config.java:92](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/oauth2/OAuth2Config.java#L92), [SharedSecretConfig.java:70](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/sharedsecret/SharedSecretConfig.java#L70)).

Keeping them matched is mostly so either backend can sit behind the same frontend/proxy without anything noticing the swap.

Happy to rename it on the JS side if you'd rather it not carry the Java-ism.

quickstart/backend-js/src/auth/oauth2.ts
fernandomg: There are two things going on here, both just OAuth2 setup for the default AppProvider rather than data seeding.

1. at startup we register the AppProvider OAuth2 client ([oauth2.ts:19](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/oauth2.ts#L19), wired in from [server.ts:19](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/server.ts#L19)) (that mirrors Java's static client registration ([application-oauth2.yml:26](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/resources/application-oauth2.yml#L26)) over its tenant baseline ([application.yml:39](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/resources/application.yml#L39)));
2. then per login (line 31) we generate the usual Authorization-Code + PKCE values: `state` for CSRF, `nonce` for ID-token replay protection, and the PKCE verifier/challenge against code interception, kept in [oauth2.ts:29-33](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/oauth2.ts#L29-L33) and checked back on the callback ([oauth2.ts:56-60](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/auth/oauth2.ts#L56-L60)). Spring handles that part under the hood in its `oauth2Login` filter ([OAuth2Config.java:83](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/security/oauth2/OAuth2Config.java#L83)); it's explicit here because we drive openid-client directly.

quickstart/backend-js/src/auth/session.ts
fernandomg: It doesn't actually have to be. `SESSION_SECRET` is optional.

When it's set we use it as a stable signing secret; when it isn't, we fall back to random bytes per boot (line 17 after `??`), which does mean a restart logs everyone out.

That's the same behavior you get from Java's in-memory HttpSession (it drops sessions on restart too), so instead of patching it on just one side we flagged it under "Deferred to follow-up" in the PR description.

quickstart/backend-js/src/domain/licensing/mappers.ts
fernandomg: Partly, yes. We do use codegen-js where it fits, like the typed command construction ([service.ts:4](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/domain/licensing/service.ts#L4)).

These mappers are a slightly different job, though: they shape the ledger payloads into our OpenAPI response and add a few derived fields, working over joined PQS rows ([mappers.ts:12](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/domain/licensing/mappers.ts#L12), [repository.ts:24](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend-js/src/domain/licensing/repository.ts#L24)).

IIUC `codegen-js` only generates the Daml-shaped types, so the REST shaping isn't something it can produce for us. The Java backend hand-writes the same mapping ([LicenseApiImpl.java:212](https://github.com/BootNodeDev/cn-quickstart/blob/143dd9eb367e5e6e49977c29e19e70c69ad29785/quickstart/backend/src/main/java/com/digitalasset/quickstart/service/LicenseApiImpl.java#L212)), so this is parity rather than a JS shortcut.

I've captured this as a "JS-only follow-up" in the description

fernandomg:
Hi @dasormeter, thanks for taking the time to review this

A bit of context that ties most of the comments together: the goal of this PR is behavioural parity with the existing Java backend, so throughout, the intent was to mirror Java's current auth/session semantics rather than introduce a new model.

That's also why the description carries a "Deferred to follow-up" section ( to flag, up front, the items that either match Java's current behaviour or need a cross-cutting fix touching both backends).

Your feedback surfaced a good one, which I've noted under a new "JS-only follow-ups" section.

PR #374, opened 2026-07-21
update splice and scribe versions

dasormeter:


dasormeter:
MANIFEST UNKOWN errors in docker compose file...

PR #378, opened 2026-08-11
fix(docs): move participant node inside optional box in high-level topology diagram

mateusz-borek-da:
## Summary
The App User's Participant node was shown outside the 'Self Hosted Option' box in the high-level architecture diagram in , which was inconsistent with the App User Infrastructure diagram where it correctly appears inside the optional box.
## Changes
- Move  inside the  ('Self Hosted Option') subgraph in the high-level diagram.
- Add  so it gets the same optional styling as the other self-hosted nodes.

### DIFF
--- a/sdk/docs/user/002-topology.md
+++ b/sdk/docs/user/002-topology.md
@@ -16,13 +16,14 @@ flowchart TD
   subgraph user [App User]
     user-browser(Browser)
     user-auth(Auth)
-    user-participant(Participant)
 
     subgraph user-host ["Self Hosted Option"]
+      user-participant(Participant)
       user-ui(UI):::customNode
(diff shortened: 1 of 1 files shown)


mateusz-borek-da:
The other thing I wasn't sure about is the first link from the Topology section in the [README](https://github.com/digital-asset/cn-quickstart/blob/main/README.md#topology). The first link there points to the documentation that is marked as deprecated. 
I understand that the deprecated doc actually describes the foundation of this project, but I wonder if we want to keep the deprecated docs as a source of truth anywhere (I found that a similar topic was covered in https://docs.canton.network/appdev/modules/m5-localnet-development)

dasormeter (dismissed):

